PR #28, opened 2022-04-27
Feat/secret mgmt

yannickstruyf3:
- Change the way credentials are handled:
     -  Allow credentials to be defined on nutanixCluster and Manager
     - Add Prism Central attribute on nutanixCluster
     - Remove endpoint, port flags from secret
     - Implement optional insecure mode
N/A
- Create CAPI cluster with new cluster-template.yml file
- Use different user/pass at manager and cluster level and check VM creation
- Use insecure false on PC with invalid certs and check failure

**Release note**:
```release-note
- Change the way credentials are handled:
     - Allow credentials to be defined on nutanixCluster and Manager
     - Add Prism Central attribute on nutanixCluster
     - Remove endpoint, port flags from secret
     - Implement optional insecure mode
```


tuxtof (dismissed):

config/rbac/role.yaml
```diff
@@ -20,8 +20,10 @@ rules:
   resources:
   - secrets
   verbs:
+  - delete
```
deepakm-ntnx, line 23: why is this needed?

deepakm-ntnx (approved):

config/rbac/role.yaml
yannickstruyf3: @deepakm-ntnx The secret is created by the cluster-template. When the nutanixcluster object is created, it will add an owner ref to the secret and also remove the secret if nutanixcluster object is deleted. In order to delete the owned secret this permission is required.

tuxtof (approved):

PR #38, opened 2022-06-01
Added project exists check at cluster level

yannickstruyf3:
Adds a checking mechanism in the NutanixCluster resource to see if a project exists. Reason is to fail the cluster creation as soon as possible if invalid project information is passed.

N/A

- Create a cluster without a project referenced -> No change in behaviour
- Create a cluster with valid project referenced -> No change in behaviour
- Create a cluster with invalid (not existing) project -> NutanixCluster will set failure message

N/A

**Release note**:
N/A

### DIFF
--- a/controllers/nutanixcluster_controller.go
+++ b/controllers/nutanixcluster_controller.go
@@ -218,7 +218,15 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
 		return reconcile.Result{}, nil
 	}
 
-	err := r.reconcileCategories(rctx)
+	err := r.reconcileProject(rctx)
+	if err != nil {
+		errorMsg := fmt.Errorf("Failed to reconcile project for cluster %s: %v", rctx.Cluster.Name, err)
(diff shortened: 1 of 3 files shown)


tuxtof (approved):

controllers/nutanixcluster_controller.go
```diff
@@ -230,6 +238,24 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
 	return reconcile.Result{}, nil
 }
 
+func (r *NutanixClusterReconciler) reconcileProject(rctx *nctx.ClusterContext) error {
```
deepakm-ntnx, line 241: we also need to modify machine controller to add teh vm to project if initially vm was not in project when cluster was created but later was added thru yaml modification

deepakm-ntnx: per @yannickstruyf3 this will be handled thru https://jira.nutanix.com/browse/KRBN-5342

deepakm-ntnx (approved):
pls see my comments above.

PR #42, opened 2022-06-10
add image info

tuxtof:

this change add image reference link at the beginning of a new release notes

image

make a new release and check the release note


.github/workflows/release.yaml
```diff
@@ -84,12 +84,24 @@ jobs:
         yq -i '(.releaseSeries[] | select(.contract == "v1beta1")).major |= env(MAJOR)' metadata.yaml
         yq -i '(.releaseSeries[] | select(.contract == "v1beta1")).minor |= env(MINOR)' metadata.yaml
 
+    - name: generate image info
```
deepakm-ntnx, line 87: should we move this just above the list of artifacts instead of at the top.

tuxtof: not possible easily, we need to entirely change the way we generate the release notes if we want to do that

deepakm-ntnx (approved):

PR #74, opened 2022-07-08
Prefix cluster name during testing with username

deepakm-ntnx:
now creating user specific name of cluster
    
    this will avoid tests running by multiple users on same setup from
    conflicting

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,8 +236,8 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 deploy: manifests kustomize prepare-local-clusterctl docker-push-kind ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	clusterctl delete --infrastructure nutanix:${LOCAL_PROVIDER_VERSION} --include-crd || true
 	clusterctl init --infrastructure nutanix:${LOCAL_PROVIDER_VERSION} -v 9
-	# cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	# $(KUSTOMIZE) build config/default | kubectl apply -f -
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default | kubectl apply -f -
 
 .PHONY: undeploy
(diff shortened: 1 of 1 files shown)


tuxtof (approved):

yannickstruyf3 (changes requested):
Let's generate a random string instead of using username

swetangfinviya (commented):
For me make print-capx-controller-logs is not working. The variable CAPX_CONTROLLER_POD_NAME is not getting created.

Here is the output:
swetang.finviya@C02X740FJHD2 cluster-api-provider-nutanix % make print-capx-controller-logs
CAPX_CONTROLLER_POD_NAME= && kubectl -n capx-system logs -f  manager
Error from server (NotFound): pods "manager" not found
make: *** [print-capx-controller-logs] Error 1

swetangfinviya (commented):
The following make command works for me:
`.PHONY: print-capx-controller-logs
print-capx-controller-logs: ## logs the controller pod output with -f mode
	kubectl -n capx-system logs -f $(shell kubectl -n capx-system get pods | grep capx-controller | awk '{print $$1}') manager`

tuxtof (approved):

Makefile
```diff
@@ -140,7 +140,7 @@ GINKGO_NOCOLOR ?= false
 FLAVOR ?= e2e
 
 TEST_NAMESPACE=capx-test-ns
-TEST_CLUSTER_NAME=mycluster
+TEST_CLUSTER_NAME=capx-cl-${USER}
```
yannickstruyf3, line 143: I would suggest to remove the user parameter and just allow the user to override TEST_CLUSTER_NAME and TEST_NAMESPACE variables

yannickstruyf3 (changes requested):

Makefile
deepakm-ntnx: could you please elaborate on your concern with usernames? 
i can use value with username as a default value but give preference to env variable if set. does that sound ok?

deepakm-ntnx:
closing it due to merge conflicts

PR #75, opened 2022-07-08
WIP: Build os image

deepakm-ntnx:
builds os image using nutanix packer plugin

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/tools/imagebuilder/terraform/scripts/build_os_image.sh
+++ b/tools/imagebuilder/terraform/scripts/build_os_image.sh
@@ -1,6 +1,11 @@
-git clone https://github.com/kubernetes-sigs/image-builder.git
+git clone https://github.com/deepakm-ntnx/image-builder.git
 export PATH=$PATH:~/.local/bin
 cd ~/image-builder/images/capi
+git checkout nutanix-imagebuilder
+
+cp ~/nutanix.json packer/nutanix/nutanix.json
+packer init ~/config.pkr.hcl
+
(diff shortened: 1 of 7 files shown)


tools/imagebuilder/terraform/scripts/build_os_image.sh
```diff
@@ -1,6 +1,11 @@
-git clone https://github.com/kubernetes-sigs/image-builder.git
+git clone https://github.com/deepakm-ntnx/image-builder.git
```
tuxtof, line 1: remove personal link

tuxtof:
we need to fully remove Terraform from the equation, with the new packer builder this is no more needed and redundant

deepakm-ntnx:
> we need to fully remove Terraform from the equation, with the new packer builder this is no more needed and redundant

In my opinion, i faced so many issues on centos building an image. python3 was required, with ssl was required. certificates were not updated. we are not using virtualenv. this change consolidates it. customers are still free to use another method

tuxtof:
> > we need to fully remove Terraform from the equation, with the new packer builder this is no more needed and redundant
> 
> In my opinion, i faced so many issues on centos building an image. python3 was required, with ssl was required. certificates were not updated. we are not using virtualenv. this change consolidates it. customers are still free to use another method

not sure to understand your problem, with the new packer builder everything is build inside a VM , and you need only one binary on your machine

PR #87, opened 2022-07-26
add filter to verify if returned cluster is PE instead of PC

yannickstruyf3:
Fix bug that occurs when PC has the same name as PE. CAPX will throw an error stating multiple PEs with the same name have been found.
This fix adds additional filtering to prevent this problem.

N/A

make test-e2e

N/A

**Release note**:
NONE

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -34,6 +34,7 @@ import (
 
 const (
 	taskSucceededMessage = "SUCCEEDED"
+	serviceNamePECluster = "AOS"
 )
 
 func CreateNutanixClient(ctx context.Context, client client.Client, nutanixCluster *infrav1.NutanixCluster) (*nutanixClientV3.Client, error) {
@@ -149,19 +150,18 @@ func findVMByName(client *nutanixClientV3.Client, vmName string) (*nutanixClient
 }
 
 func getPEUUID(client *nutanixClientV3.Client, peName, peUUID *string) (string, error) {
-	var foundPEUUID string
 	if peUUID == nil && peName == nil {
 		return "", fmt.Errorf("cluster name or uuid must be passed in order to retrieve the pe")
 	}
-	if peUUID != nil {
+	if peUUID != nil && *peUUID != "" {
 		peIntentResponse, err := client.V3.GetCluster(*peUUID)
 		if err != nil {
 			if strings.Contains(fmt.Sprint(err), "ENTITY_NOT_FOUND") {
 				return "", fmt.Errorf("failed to find Prism Element cluster with UUID %s: %v", *peUUID, err)
 			}
 		}
-		foundPEUUID = *peIntentResponse.Metadata.UUID
-	} else if peName != nil {
+		return *peIntentResponse.Metadata.UUID, nil
+	} else if peName != nil && *peName != "" {
 		filter := getFilterForName(*peName)
 		responsePEs, err := client.V3.ListAllCluster(filter)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


tuxtof (dismissed):
LGTM

controllers/helpers.go
```diff
@@ -582,3 +584,18 @@ func getSubnetClientSideFilter(peUUID string) []*nutanixClient.AdditionalFilter
...
+
+func hasPEClusterServiceEnabled(peCluster *nutanixClientV3.ClusterIntentResponse, serviceName string) bool {
+	if peCluster.Status == nil ||
+		peCluster.Status.Resources == nil ||
+		peCluster.Status.Resources.Config == nil {
+		return false
+	}
+	serviceList := peCluster.Status.Resources.Config.ServiceList
+	for _, s := range serviceList {
+		if strings.ToUpper(*s) == serviceName {
```
thunderboltsid, line 596: don't deref a pointer without nil-checking first else it'll cause the process to panic.

controllers/helpers.go
```diff
@@ -182,10 +183,11 @@ func getPEUUID(client *nutanixClientV3.Client, peName, peUUID *string) (string,
...
 			foundPEUUID = *foundPEs[0].Metadata.UUID
 		}
-		if foundPEUUID == "" {
-			return "", fmt.Errorf("failed to retrieve Prism Element cluster by name or uuid. Verify input parameters.")
-		}
 	}
+	if foundPEUUID == "" {
+		return "", fmt.Errorf("failed to retrieve Prism Element cluster by name or uuid. Verify input parameters.")
+	}
+
```
thunderboltsid, line 190: Adding a comment here because it won't let me at the function declaration. My understanding is that the function is supposed to return the PE UUID given a name or a UUID. This smells; why is the UUID a parameter?

tuxtof (dismissed):

controllers/helpers.go
thunderboltsid: something like this is a lot cleaner
```go
func getPrismelement(client *nutanixClientV3.Client, peName, peUUID string) (string, error) {
	if peUUID != "" {
	    // UUID already exists just verify if it corresponds to a real PE
	    peIntentResponse, err := client.V3.GetCluster(peUUID)
		if err != nil {
			if strings.Contains(fmt.Sprint(err), "ENTITY_NOT_FOUND") {
				return "", fmt.Errorf("failed to find Prism Element cluster with UUID %s: %v", *peUUID, err)
			}
		}
		return *peIntentResponse.Metadata.UUID, nil
	}
	if peName != "" {
		filter := getFilterForName(peName)
		responsePEs, err := client.V3.ListAllCluster(filter)
		if err != nil {
			return "", err
		}
		// Validate filtered PEs
		foundPEs := make([]*nutanixClientV3.ClusterIntentResponse, 0)
		for _, s := range responsePEs.Entities {
			peSpec := s.Spec
			if *peSpec.Name == peName && hasPEClusterServiceEnabled(s, serviceNamePECluster) {
				foundPEs = append(foundPEs, s)
			}
		}
		if len(foundPEs) == 1 {
			return *foundPEs[0].Metadata.UUID
		} else if len(foundPEs) == 0 {
			return "", fmt.Errorf("failed to retrieve Prism Element cluster by name %s", peName)
		} else {
			return "", fmt.Errorf("more than one Prism Element cluster found with name %s", peName)
		}
	}

	return "", fmt.Errorf("failed to retrieve Prism Element cluster by name %s or uuid %s", peName, peUUID)
}
```

controllers/helpers.go
```diff
@@ -149,7 +150,6 @@ func findVMByName(client *nutanixClientV3.Client, vmName string) (*nutanixClient
 }
 
 func getPEUUID(client *nutanixClientV3.Client, peName, peUUID *string) (string, error) {
```
thunderboltsid, line 152: note: the pointers might not be nil by the value could be a zero value so consider checking for both.

thunderboltsid (approved):

tuxtof (approved):

PR #106, opened 2022-08-11
fix flavor patch

tuxtof:

Change the patch method to avoid objects override

`make cluster-templates` and diff flavors result with base to see if nothing was removed

**Release note**:
```release-note
NONE
```


yannickstruyf3 (approved):

vnephologist (approved):
Lgtm

PR #183, opened 2022-12-08
fix base64 line wrapping

tuxtof:

fix base64 value for CSI cert gen script on linux platform

launch the gen-self-cert.sh script on the following platform and verify variable are correctly encoded
- MacOS X (13.0)
- Centos 7.9
- Ubuntu 20.04

**Release note**:
```release-note
fix base64 value for CSI cert gen script on linux platform
```


vnephologist (approved):
Lgtm

PR #190, opened 2022-12-13
Add uuid cluster creation testing and capxhelpers

yannickstruyf3:
- Adds e2e test for testing cluster creation via UUIDs
- Adds Helpers interface for CAPX functions

N/A

CAPX feature e2e:
```
Ran 14 of 25 Specs in 1917.475 seconds
SUCCESS! -- 14 Passed | 0 Failed | 0 Pending | 11 Skipped
PASS
```

N/A

**Release note**:

```release-note
- Add e2e test for testing cluster creation via UUIDs
- Adds Helpers interface for CAPX functions
```

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -40,32 +40,8 @@ const (
 	serviceNamePECluster = "AOS"
 )
 
-type Helpers interface {
-	CreateNutanixClient(secretInformer coreinformers.SecretInformer, cmInformer coreinformers.ConfigMapInformer, nutanixCluster *infrav1.NutanixCluster) (*nutanixClientV3.Client, error)
-	CreateSystemDiskSpec(imageUUID string, systemDiskSize int64) (*nutanixClientV3.VMDisk, error)
-	DeleteCategories(ctx context.Context, client *nutanixClientV3.Client, categoryIdentifiers []*infrav1.NutanixCategoryIdentifier) error
-	DeleteVM(ctx context.Context, client *nutanixClientV3.Client, vmName, vmUUID string) (string, error)
-	FindVM(ctx context.Context, client *nutanixClientV3.Client, nutanixMachine *infrav1.NutanixMachine) (*nutanixClientV3.VMIntentResponse, error)
-	FindVMByName(ctx context.Context, client *nutanixClientV3.Client, vmName string) (*nutanixClientV3.VMIntentResponse, error)
-	FindVMByUUID(ctx context.Context, client *nutanixClientV3.Client, uuid string) (*nutanixClientV3.VMIntentResponse, error)
-	GenerateProviderID(uuid string) string
-	GetCategoryVMSpec(ctx context.Context, client *nutanixClientV3.Client, categoryIdentifiers []*infrav1.NutanixCategoryIdentifier) (map[string]string, error)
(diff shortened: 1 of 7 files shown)


tuxtof (approved):
LGTM

controllers/helpers.go
```diff
@@ -40,7 +40,32 @@ const (
 	serviceNamePECluster = "AOS"
 )
 
-func CreateNutanixClient(secretInformer coreinformers.SecretInformer, cmInformer coreinformers.ConfigMapInformer, nutanixCluster *infrav1.NutanixCluster) (*nutanixClientV3.Client, error) {
+type Helpers interface {
```
thunderboltsid, line 43: I can't see any real reason for this interface. IMO, this straight-up abuses the concept of interfaces. Let's remove this; If needed, in the future, we can add more precise interfaces for mocking instead of one large helpers interface that realistically will never see another implementation of the interface.

thunderboltsid: cc @deepakm-ntnx

deepakm-ntnx: @thunderboltsid am not seeing any good reason in your description as to why a large interface is not good vs why we need smaller. could you please elaborate? basically we can mock all these function using same mockgen created code. no? cc @yannickstruyf3

thunderboltsid: In Go interfaces are a mechanism to create abstractions that can be fulfilled using multiple implementations. What you're suggesting creating an abstraction for sake of creating an abstraction and is considered interface pollution in Go world. As for the large interface, in Go (unlike Java, et al), larger interfaces are generally frowned upon (The larger the interface; the weaker the abstraction): https://www.youtube.com/watch?v=PAAkCSZUG1c&t=317s.

deepakm-ntnx: - Controllers/helpers.go had global wrapper functions which act as an intermediate function to call actual client function with capx specific logic.
- Ideally this should not be called as helpers interface but a NutanixClientAPIWrapper
- Moreover, the struct we use should encapsulate client connection so that we can make this a complete object for mocking. 
- Then we can mock these function to test client as well as our custom logic in it.
- I don't think we should get rid of this interface but make the implemented struct more complete so that we can mock these function in unit tests and have 100% of code coverage.
- I am ok to reduce the interface size by breaking it down into smaller but getting rid of it when we have to add unit tests to it is not right in my opinion. Unless you have a better solution/proposal for unit tests without mocking these?

thunderboltsid: Two separate points:
1. The interface shouldn't be added to this PR. To start with, there is no good reason why the interface needs to be added to this PR. The PR adds an e2e test, and it should focus on that. As for your argument on the interface needing to be in the PR for unit testing, I don't see a single unit test in the PR. Generally, any refactoring to add an interface can be done in a separate future PR.
2. The interface itself is debatable. I'm open to discussing this but only in the context of another future PR. we can add an interface for items that genuinely warrant having an interface (i.e., things that require an external call: the mock should be on the Prism API call and not on its wrapper: wrapper is business logic and should be part of unit test execution and coverage). By mocking our own helper functions, we are refusing to test them which doesn't make much sense to me.

I argue that we let this PR add the e2e test and have a separate PR handle any refactoring for unit testing/interfaces. If you want, I can undertake the responsibility of creating that PR.

tuxtof (approved):

thunderboltsid (approved):

PR #191, opened 2022-12-13
changed reconcileNode logic to not raise error for expected behaviour

yannickstruyf3:
CAPX raises errors for expected behaviour. This PR makes sure the reconciliation does not receive a failure if a node is not yet ready or Kubernetes API is not available during provisioning. 

N/A

CAPX feature test: 
```
Ran 12 of 23 Specs in 1501.631 seconds
SUCCESS! -- 12 Passed | 0 Failed | 0 Pending | 11 Skipped
PASS
```

N/A

**Release note**:

```release-note
- Change reconcileNode logic to handle expected behaviour instead of returning errors
```

### DIFF
--- a/controllers/nutanixmachine_controller.go
+++ b/controllers/nutanixmachine_controller.go
@@ -347,10 +347,13 @@ func (r *NutanixMachineReconciler) reconcileNormal(rctx *nctx.MachineContext) (r
 		klog.Infof("%s The NutanixMachine is already ready, providerID: %s", rctx.LogPrefix, rctx.NutanixMachine.Spec.ProviderID)
 
 		if rctx.NutanixMachine.Status.NodeRef == nil {
-			err = r.reconcileNode(rctx)
+			reconcileNodeResult, err := r.reconcileNode(rctx)
 			if err != nil {
 				klog.Errorf("%s Failed to reconcile the workload cluster node. %v", rctx.LogPrefix, err)
 				return reconcile.Result{}, err
+			}
+			if reconcileNodeResult != nil {
+				return *reconcileNodeResult, nil
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


tuxtof (approved):

controllers/nutanixmachine_controller.go
```diff
@@ -475,23 +480,23 @@ func (r *NutanixMachineReconciler) reconcileNode(rctx *nctx.MachineContext) erro
...
 	err = patchHelper.Patch(rctx.Context, node)
 	if err != nil {
 		klog.Errorf("%s Failed to patch the remote workload cluster node %s's spec.providerID. %v", rctx.LogPrefix, nodeName, err)
-		return err
+		return nil, err
 	}
 	klog.Infof("%s Patched the workload node %s spec.providerID: %s", rctx.LogPrefix, nodeName, node.Spec.ProviderID)
 
-	return nil
+	return nil, nil
```
thunderboltsid, line 494: imo, this return a `reconcile.Result{}, nil` ?

controllers/nutanixmachine_controller.go
```diff
@@ -856,3 +861,8 @@ func (r *NutanixMachineReconciler) addVMToProject(rctx *nctx.MachineContext, vmM
 	conditions.MarkTrue(rctx.NutanixMachine, infrav1.ProjectAssignedCondition)
 	return nil
 }
+
+func (r *NutanixMachineReconciler) isGetRemoteClientConnectionError(err error) bool {
+	const expectedErrString = "connect: connection refused"
```
deepakm-ntnx, line 866: pls add comment stating possible reasons of this error

controllers/nutanixmachine_controller.go
yannickstruyf3: We should not return a reconcile.Result since we don't want to trigger a new reconciliation loop. Hence it is nil.

controllers/nutanixmachine_controller.go
yannickstruyf3: Added comment

tuxtof (approved):

thunderboltsid (approved):

PR #230, opened 2023-02-23
Added logic to filter vlan and overlay subnets

yannickstruyf3:
Fixes issue that prevents CAPX clusters to be deployed inside VPCs. This fix will add logic to determine if a subnet is overlay or not.

N/A

Manual testing in a VPC
e2e tests:
```
Ran 14 of 26 Specs in 2769.665 seconds
SUCCESS! -- 14 Passed | 0 Failed | 0 Pending | 12 Skipped
PASS
```

N/A

**Release note**:

```release-note
- Allow CAPX clusters to be deployed inside VPCs
```

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -25,7 +25,6 @@ import (
 	"github.com/google/uuid"
 	infrav1 "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/api/v1beta1"
 	nutanixClientHelper "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkg/client"
-	nutanixClient "github.com/nutanix-cloud-native/prism-go-client"
 	"github.com/nutanix-cloud-native/prism-go-client/utils"
 	nutanixClientV3 "github.com/nutanix-cloud-native/prism-go-client/v3"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -38,6 +37,8 @@ const (
 
 	taskSucceededMessage = "SUCCEEDED"
 	serviceNamePECluster = "AOS"
+
+	subnetTypeOverlay = "OVERLAY"
 )
 
 // CreateNutanixClient creates a new Nutanix client from the environment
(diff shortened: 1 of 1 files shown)


controllers/helpers.go
```diff
@@ -256,17 +257,26 @@ func GetSubnetUUID(ctx context.Context, client *nutanixClientV3.Client, peUUID s
...
 		if err != nil {
 			return "", err
 		}
 		// Validate filtered Subnets
 		foundSubnets := make([]*nutanixClientV3.SubnetIntentResponse, 0)
 		for _, s := range responseSubnets.Entities {
 			subnetSpec := s.Spec
-			if *subnetSpec.Name == *subnetName && *subnetSpec.ClusterReference.UUID == peUUID {
-				foundSubnets = append(foundSubnets, s)
+			if *subnetSpec.Name == *subnetName && subnetSpec.Resources != nil && subnetSpec.Resources.SubnetType != nil {
```
thunderboltsid, line 269: before deref-ing a pointer we should always add nil-checks in place.

controllers/helpers.go
```diff
@@ -256,17 +257,26 @@ func GetSubnetUUID(ctx context.Context, client *nutanixClientV3.Client, peUUID s
...
+				switch *subnetSpec.Resources.SubnetType {
+				case subnetTypeOverlay:
+					// Overlay subnets are present on all PEs managed by PC.
+					foundSubnets = append(foundSubnets, s)
+				// By default check if the PE UUID matches if it is not an overlay subnet.
+				default:
+					if *subnetSpec.ClusterReference.UUID == peUUID {
+						foundSubnets = append(foundSubnets, s)
+					}
+				}
```
thunderboltsid, line 279: given it's just two types we deal with atm, we should perhaps just use a `if` conditional here.

controllers/helpers.go
```diff
@@ -256,17 +257,26 @@ func GetSubnetUUID(ctx context.Context, client *nutanixClientV3.Client, peUUID s
 		foundSubnetUUID = *subnetIntentResponse.Metadata.UUID
 	} else if subnetName != nil {
 		filter := getFilterForName(*subnetName)
-		subnetClientSideFilter := getSubnetClientSideFilter(peUUID)
-		responseSubnets, err := client.V3.ListAllSubnet(ctx, filter, subnetClientSideFilter)
+		// Not using additional filtering since we want to list overlay and vlan subnets
```
thunderboltsid, line 260: this is a good idea specially because server side resource filtering is a bit hit & miss in my experience and doesn't always work as expected.

controllers/helpers.go
yannickstruyf3: Fixed

controllers/helpers.go
yannickstruyf3: Fixed

thunderboltsid (approved):

deepakm-ntnx (approved):

PR #301, opened 2023-08-15
Fix credentials delete logic

adiantum:
Fixes race condition on cluster delete if credentials secret attempted to delete before the cluster:

 - check if no delete attempt occurred
 - check if no secret present in cluster
 - add integration test case in e2e suite
 - change Makefile test-e2e target to make able to run selected tests using GINKGO_FOCUS env variable

Fix issues: Cluster and credentials issue race condition

How Has This Been Tested?:
E2E test case added.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/controllers/nutanixcluster_controller.go
+++ b/controllers/nutanixcluster_controller.go
@@ -321,6 +321,10 @@ func (r *NutanixClusterReconciler) reconcileCredentialRefDelete(ctx context.Cont
 	}
 	err = r.Client.Get(ctx, secretKey, secret)
 	if err != nil {
+		if errors.IsNotFound(err) {
+			log.V(1).Info(fmt.Sprintf("Secret %s in namespace %s for cluster %s not found. Ignoring since object must be deleted", secret.Name, secret.Namespace, nutanixCluster.Name))
+			return nil
+		}
 		return err
 	}
 	ctrlutil.RemoveFinalizer(secret, infrav1.NutanixClusterCredentialFinalizer)
@@ -328,10 +332,14 @@ func (r *NutanixClusterReconciler) reconcileCredentialRefDelete(ctx context.Cont
 	if err := r.Client.Update(ctx, secret); err != nil {
 		return err
 	}
-	log.Info(fmt.Sprintf("removing secret %s in namespace %s for cluster %s", secret.Name, secret.Namespace, nutanixCluster.Name))
-	if err := r.Client.Delete(ctx, secret); err != nil {
-		return err
+
+	if secret.DeletionTimestamp.IsZero() {
+		log.Info(fmt.Sprintf("removing secret %s in namespace %s for cluster %s", secret.Name, secret.Namespace, nutanixCluster.Name))
+		if err := r.Client.Delete(ctx, secret); err != nil {
+			return err
+		}
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


controllers/nutanixcluster_controller.go
```diff
@@ -308,17 +308,25 @@ func (r *NutanixClusterReconciler) reconcileCredentialRefDelete(ctx context.Cont
...
 	if err := r.Client.Update(ctx, secret); err != nil {
 		return err
 	}
-	log.Info(fmt.Sprintf("removing secret %s in namespace %s for cluster %s", secret.Name, secret.Namespace, nutanixCluster.Name))
-	if err := r.Client.Delete(ctx, secret); err != nil {
-		return err
+
+	if secret.DeletionTimestamp.IsZero() {
+		log.Info(fmt.Sprintf("removing secret %s in namespace %s for cluster %s", secret.Name, secret.Namespace, nutanixCluster.Name))
+		if err := r.Client.Delete(ctx, secret); err != nil {
```
deepakm-ntnx, line 325: should we update the conditions too with this state

deepakm-ntnx: should we be deleting secret by ourselves? per https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/ see eg of pv-protection. we are not the owner controller of secret resource cc @yannickstruyf3

deepakm-ntnx: I dont think we should be calling delete on secret in our operator. https://martinheinz.dev/blog/74 we should remove finalizer, update it on secret and continue. let the operator which handles secret object take care of it. we should not call delete inside deletetimestamp = 0 case as its already being deleted anyway

yannickstruyf3: Remove the secret deletion for now and only cleanup finalizer

adiantum: Other CAPI providers usually delete the secret (if using a secret):
 - CAPV: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/controllers/vspherecluster_reconciler.go#L213
 - CAPH: https://github.com/syself/cluster-api-provider-hetzner/blob/1b1b4c55071aaf9001317649df23e585118e4478/controllers/hetznercluster_controller.go#L264C34-L264C34

Also in CAPI test framework there is expectation that after cluster delete will be no resources in namespace:
https://github.com/kubernetes-sigs/cluster-api/blob/main/test/framework/cluster_helpers.go#L237

test/e2e/capx_regression.go
```diff
@@ -0,0 +1,215 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	infrav1 "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/api/v1beta1"
+)
+
+var _ = Describe("Nutanix regression tests", Label("capx-regression-test", "regression", "slow", "network"), func() {
+	const (
+		specName = "capx-regression"
+
+		controlplaneEndpointIPKey       = "CONTROL_PLANE_ENDPOINT_IP"
```
yannickstruyf3, line 43: Can we move these constants to a shared file? Isn't this one used in a different test as well?

yannickstruyf3 (changes requested):

tuxtof (approved):
LGTM

PR #328, opened 2023-11-10
Add codecov.yml to ensure adequate patch coverage

thunderboltsid:
This sets the minimum patch coverage required for changes to repo to 75%.

There was a discrepancy between the Go version in go.mod
and the go version in github actions workflow. This change also
alleviates that discrepancy by changing the Go version in both to
the latest Go version (v1.21).

**How has this been tested?**
Screenshot 2023-11-11 at 01 18 20

PR #331, opened 2023-11-20
Make CCM for CAPX mandatory by default

adiantum:
Make CCM component mandatory default for CAPX deployment.
 - get rid of node reconcile
 - changes manifests for CAPX deployment
 - add test for CAPX provider update and changes Makefile

Current node provider id logic isn't optimal and contains drawbacks. After discussion we've decided to use CCM as state of the art solution that helps us to improve known drawbacks and maintain mainstream solution in future.

Tested with clusterctl-upgrade tests for provider version 1.2.4 => upstream and for K8S 1.26&1.27.
The test upgrade CAPX version from older releases to the latest one and apply required CCM manifests.

**Release note**:
```release-note
IMPORTANT: Since this version CCM will be installed with CAPX by default and become mandatory. If you update CAPX from previous version you should also install CCM and perform migration actions. Please see documentation for details.
```


yannickstruyf3 (approved):

deepakm-ntnx:
Since this is a big change, pls add more details in the description explaining the feature and general workflows and reasons why this is required

Makefile
```diff
@@ -16,7 +16,7 @@ IMG_TAG := latest
 endif
 
 ifeq (${LOCAL_PROVIDER_VERSION},latest)
-LOCAL_PROVIDER_VERSION := v0.0.0
+LOCAL_PROVIDER_VERSION := v1.3.99
```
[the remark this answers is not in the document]
adiantum: Yes, to test upgrade between CAPX version we need upstream version higher than released.

deepakm-ntnx:
@thunderboltsid @adiantum why is codecov is not showing up here? does it need to be sync'd with latest master since this was added then [ttps://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/328](https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/328)

Makefile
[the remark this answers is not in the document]
deepakm-ntnx: pls add comment and when to update it here in code

yannickstruyf3 (dismissed):

tuxtof (dismissed):
LGTM

Makefile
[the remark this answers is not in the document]
adiantum: Fixed

controllers/nutanixcluster_controller.go
```diff
@@ -360,9 +362,8 @@ func (r *NutanixClusterReconciler) reconcileCredentialRefDelete(ctx context.Cont
 
 	if secret.DeletionTimestamp.IsZero() {
 		log.Info(fmt.Sprintf("removing secret %s in namespace %s for cluster %s", secret.Name, secret.Namespace, nutanixCluster.Name))
-		if err := r.Client.Delete(ctx, secret); err != nil {
-			return err
-		}
+		// If the secret is already deleted, don't return error
+		r.Client.Delete(ctx, secret)
```
yannickstruyf3, line 366: Why was this changed in this PR? Were there issues while testing?

controllers/nutanixcluster_controller_test.go
```diff
@@ -216,4 +217,172 @@ func TestNutanixClusterReconciler(t *testing.T) {
...
+					Expect(err).NotTo(HaveOccurred())
+				}()
+
+				// Reconile Delete credential ref
+				err := reconciler.reconcileCredentialRefDelete(ctx, ntnxCluster)
+				g.Expect(err).To(HaveOccurred())
+			})
+		})
+
+		Context("Delete credentials ref reconcile failed: credential ref is null", func() {
```
yannickstruyf3, line 318: The Context line "cred ref is null" doesn't seem to match the test case because credential Ref is set on the ntnxCluster.

adiantum: Fixed

yannickstruyf3 (dismissed):

Makefile
```diff
@@ -132,6 +133,15 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Get latest git hash
+GIT_COMMIT_HASH=$(shell git rev-parse HEAD)
+
+# Set E2E_IMG_TAG to override the default image tag used for e2e tests
+E2E_IMG_TAG="e2e-${GIT_COMMIT_HASH}"
```
tuxtof, line 140: let's reuse the existing IMG_TAG

Makefile
```diff
@@ -327,15 +338,14 @@ cluster-e2e-templates-no-kubeproxy: $(KUSTOMIZE) ##Generate cluster templates wi
...
 ##@ Testing
 
 .PHONY: docker-build-e2e
 docker-build-e2e: $(KO) ## Build docker image with the manager with e2e tag.
-	GIT_COMMIT_HASH=`git rev-parse HEAD` && \
-	KO_DOCKER_REPO=ko.local GOFLAGS="-ldflags=-X=main.gitCommitHash=$${GIT_COMMIT_HASH}" $(KO) build -B --platform=${PLATFORMS_E2E} -t e2e -L .
-	docker tag ko.local/cluster-api-provider-nutanix:e2e ${IMG_REPO}:e2e
+	echo "Git commit hash: ${GIT_COMMIT_HASH}"
+	KO_DOCKER_REPO=ko.local GOFLAGS="-ldflags=-X=main.gitCommitHash=${GIT_COMMIT_HASH}" $(KO) build -B --platform=${PLATFORMS_E2E} -t e2e-${GIT_COMMIT_HASH} -L .
+	docker tag ko.local/cluster-api-provider-nutanix:e2e-${GIT_COMMIT_HASH} ${IMG_REPO}:e2e
```
tuxtof, line 348: let's use the E2E_IMG_TAG for the target tag

Makefile
```diff
@@ -132,6 +133,15 @@ else
...
 endif
 
+# Get latest git hash
+GIT_COMMIT_HASH=$(shell git rev-parse HEAD)
+
+# Set E2E_IMG_TAG to override the default image tag used for e2e tests
+E2E_IMG_TAG="e2e-${GIT_COMMIT_HASH}"
+
+# Set docker image repository for E2E clusterctl update tests
+HARBOR_REPO?="harbor"
```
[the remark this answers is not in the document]
adiantum: fixed

Makefile
adiantum: fixed

Makefile
adiantum: fixed

Makefile
```diff
@@ -468,6 +478,12 @@ test-e2e-cilium-no-kubeproxy:
 .PHONY: test-e2e-all-cni
 test-e2e-all-cni: test-e2e test-e2e-calico test-e2e-flannel test-e2e-cilium test-e2e-cilium-no-kubeproxy
 
+.PHONY: test-e2e-clusterctl-upgrade
+test-e2e-clusterctl-upgrade: docker-build-e2e $(GINKGO_BIN) cluster-e2e-templates cluster-templates ## Run the end-to-end tests
+	echo "Image tag for E2E test is ${E2E_IMG_TAG}"
+	docker tag ko.local/cluster-api-provider-nutanix:${E2E_IMG_TAG} ${HARBOR_REPO}/controller:${E2E_IMG_TAG}
+	docker push ${HARBOR_REPO}/controller:${E2E_IMG_TAG}
```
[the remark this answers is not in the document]
adiantum: fixed

controllers/nutanixmachine_controller.go
```diff
@@ -445,73 +441,6 @@ func (r *NutanixMachineReconciler) reconcileNormal(rctx *nctx.MachineContext) (r
...
-	}
-
-	// Update the node's Spec.ProviderID
-	patchHelper, err := patch.NewHelper(node, remoteClient)
-	if err != nil {
-		log.Error(err, fmt.Sprintf("failed to create patchHelper for the workload cluster node %s", nodeName))
-		return reconcile.Result{}, err
-	}
-
-	node.Spec.ProviderID = rctx.NutanixMachine.Spec.ProviderID
```
deepakm-ntnx, line 504: who will set this if this code is removed?

adiantum: CCM will handle this.

test/e2e/config/nutanix.yaml
```diff
@@ -13,6 +13,12 @@ images:
     loadBehavior: mustLoad
   # ## PLEASE KEEP THESE UP TO DATE WITH THE COMPONENTS
   # Cluster API v1beta1 Preloads
+  - name: registry.k8s.io/cluster-api/cluster-api-controller:v1.3.5
```
thunderboltsid, line 16: Let's change controller versions we pre-fetch to v1.4.9, v1.5.4, and v1.6.0

thunderboltsid: https://github.com/kubernetes-sigs/cluster-api/releases

thunderboltsid: Note: We can take this in a separate PR.

test/e2e/log/log.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func Errorf(format string, a ...interface{}) {
+	Logf(LogError, format, a...)
+}
+
+func Logf(level string, format string, a ...interface{}) {
+	msg := level + ": " + format + "\n"
+	fmt.Fprintf(GinkgoWriter, msg, a...)
+}
```
thunderboltsid, line 54: Add a comment above all exported funcs and consts.

test/e2e/config/nutanix.yaml
adiantum: Done

test/e2e/log/log.go
adiantum: Done

controllers/nutanixcluster_controller.go
```diff
@@ -333,9 +333,11 @@ func (r *NutanixClusterReconciler) reconcileCredentialRefDelete(ctx context.Cont
 	log := ctrl.LoggerFrom(ctx)
 	credentialRef, err := nutanixClient.GetCredentialRefForCluster(nutanixCluster)
 	if err != nil {
+		log.Error(err, fmt.Sprintf("error occurred while getting credential ref for cluster %s", nutanixCluster.Name))
```
jimmidyson, line 336: Please use structured logging (guidelines in https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/TMP-LOGGING.md?plain=1#L31 are useful btw).
```suggestion
		log = log.WithValues("cluster", client.ObjectKeyFromObject(cluster))
		log.Error(err, fmt.Sprintf("error occurred while getting credential ref for cluster")
```

My preference is to initialize the logger in the original `Reconcile` func like this to ensure all logs contain the object being reconciled:

```go
log := ctrl.LoggerFrom(ctx).WithValues("cluster", req.NamespacedName)
```

controllers/nutanixcluster_controller.go
jimmidyson, line 366: This should check the error is a not found error:

```suggestion
		if err := r.Client.Delete(ctx, secret); err != nil && !errors.IsNotFound(err) {
			return err
		}
```

jimmidyson: As a follow-up, why can't this be handled by ownership and GC? Is this code even needed?

yannickstruyf3 (approved):

hack/install-go.sh
```diff
@@ -4,8 +4,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-wget https://go.dev/dl/go1.21.4.linux-amd64.tar.gz
+wget -q https://go.dev/dl/go1.21.4.linux-amd64.tar.gz
 rm -rf /usr/local/go && tar -C /usr/local -xzf go1.21.4.linux-amd64.tar.gz
```
jimmidyson, line 8: In future, can we not update go for the developer user machine like this? Many users have set up go for various projects and an individual project should not potentially globally change the installed go version, which could affect other projects builds. I think this script may only be applicable in CI so maybe this is OK, but generally I prefer tooling to be project specific, I've hit too many cross-project tooling issues to like projects managing global tooling like this.

test/e2e/config/nutanix.yaml
```diff
@@ -13,12 +13,31 @@ images:
     loadBehavior: mustLoad
   # ## PLEASE KEEP THESE UP TO DATE WITH THE COMPONENTS
   # Cluster API v1beta1 Preloads
-  - name: registry.k8s.io/cluster-api/cluster-api-controller:v1.4.1
+  - name: registry.k8s.io/cluster-api/cluster-api-controller:v1.3.10
```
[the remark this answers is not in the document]
thunderboltsid: They are not (see the full change). They represent the list of latest CAPI controllers for each minor version.

hack/install-go.sh
thunderboltsid: Sure, in future, we can consider using devbox and the like to have a virtual environment that's project specific. Although, that'd be out of scope for this PR.

Makefile
```diff
@@ -16,7 +16,8 @@ IMG_TAG := latest
 endif
 
 ifeq (${LOCAL_PROVIDER_VERSION},latest)
-LOCAL_PROVIDER_VERSION := v0.0.0
+# Change this versions after release when required here and in e2e config (test/e2e/config/nutanix.yaml)
+LOCAL_PROVIDER_VERSION := v1.3.99
```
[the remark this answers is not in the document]
thunderboltsid: That's a nifty utility :D

tuxtof (dismissed):
LGTM

Makefile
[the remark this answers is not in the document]
adiantum: @jimmidyson the tools looks pretty good, but to use it with env variables we need to refactor e2e test scripts. I really would prefer to make it in separate PR.

test/e2e/config/nutanix.yaml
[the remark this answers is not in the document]
adiantum: @jimmidyson In fact it's upgrade with additional versions. Please take a look one more time the full changes with extended context. @thunderboltsid asked me to add more preloads here: https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/331#discussion_r1422434589

Makefile
```diff
@@ -132,6 +133,21 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
+# Get latest git hash
+GIT_COMMIT_HASH=$(shell git rev-parse HEAD)
+
+# Get the local image registry required for clusterctl upgrade tests
+LOCAL_IMAGE_REGISTRY ?= localhost:5000
```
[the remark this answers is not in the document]
adiantum: @jimmidyson Yes, we really need image registry that will be available from test Nutanix cluster. 
To test CAPX controller upgrade I've reused clusterctl-upgrade test from CAPI: https://github.com/kubernetes-sigs/cluster-api/blob/v1.3.5/test/e2e/clusterctl_upgrade.go#L209 This test creates bootstrap cluster (kind) next it creates workload cluster on bootstrap and use `cluster init` to make it (workload cluster) the management cluster with old CAPX provider version, then on the mgmt cluster test creates workload cluster. So in fact we have 3 clusters: 1. bootstrap(kind) => 2. management => 3. workload. The management cluster during upgrade requires the latest (e2e) image build which it can not take from ko.local (this cluster is not local kind cluster, rather the cluster on Nutanix test cluster).
In CAPI they recommend to use AWS S3 and docker save to make latest build available for management cluster: https://github.com/kubernetes-sigs/cluster-api/blob/v1.3.5/test/e2e/clusterctl_upgrade.go#L136 But I don't like this approach because it means patch to manifests with additional preKubeadmCommands. So, to solve it I utilize the local harbor using above variable.

Makefile
```diff
@@ -132,6 +133,21 @@ else
...
+
+ifeq (${MAKECMDGOALS},test-e2e-clusterctl-upgrade)
+	IMG_TAG=e2e-${GIT_COMMIT_HASH}
+	IMG_REPO=${LOCAL_IMAGE_REGISTRY}/controller
+endif
+
+ifeq (${MAKECMDGOALS},docker-build-e2e)
+	IMG_TAG=e2e-${GIT_COMMIT_HASH}
+endif
```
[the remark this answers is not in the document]
adiantum: In fact we don't need all dependent targets to inherit these values. I used to use separate variables, but @tuxtof asked me to use IMG_TAG and IMG_REPO: https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/331#discussion_r1419102887 and https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/331#discussion_r1419105460

Makefile
```diff
@@ -285,7 +301,10 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 ##@ Templates
 
 .PHONY: cluster-e2e-templates
-cluster-e2e-templates: $(KUSTOMIZE) cluster-e2e-templates-v1beta1 cluster-e2e-templates-v1alpha4 ## Generate cluster templates for all versions
+cluster-e2e-templates: $(KUSTOMIZE) cluster-e2e-templates-v1beta1 cluster-e2e-templates-v1alpha4 cluster-e2e-templates-v124 ## Generate cluster templates for all versions
+
+cluster-e2e-templates-v124: $(KUSTOMIZE) ## Generate cluster templates for CAPX v1.2.4
```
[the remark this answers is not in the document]
adiantum: @jimmidyson This template used in E2EConfig. Maybe you right and it's better to download template from release, but I also prefer to make these changes in separate PR.

controllers/nutanixcluster_controller.go
adiantum: @jimmidyson Structured logging is a great thing, but we need to change it not only in this line. Let's make logging improvement in separate PR.

controllers/nutanixcluster_controller.go
adiantum: The secret created with separate manifest and set ownership during reconciliation.
Also we planned to refactor these part to improve testability. Right know there is limited ability to test race conditions. I think it's better to improve in separate PR.

hack/install-go.sh
adiantum: Let's make it in separate PR.

Makefile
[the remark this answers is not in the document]
jimmidyson: @adiantum Makes sense to do as follow up PR 👍

Makefile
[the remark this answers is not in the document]
jimmidyson: That makes sense, thanks for explaining.

Makefile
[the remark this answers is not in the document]
jimmidyson: But won't setting them here make all dependent targets use these anyway?

Makefile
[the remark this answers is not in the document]
jimmidyson: As long as this doesn't get missed, I'm happy with a follow up PR. My reason for wanting to use release artifacts is so these old versions don't get mutated in this repo and skew e2e test results.

controllers/nutanixcluster_controller.go
jimmidyson: 👍

controllers/nutanixcluster_controller.go
jimmidyson: OK, but let's at least add the improved error checking in this PR.

hack/install-go.sh
jimmidyson: 👍

test/e2e/config/nutanix.yaml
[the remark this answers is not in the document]
jimmidyson: Ah yes, sorry misread this change.

Makefile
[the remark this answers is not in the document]
jimmidyson: Agreed :)

Makefile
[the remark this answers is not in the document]
jimmidyson: Thanks for the explanation! Makes sense to me!

Makefile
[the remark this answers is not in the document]
jimmidyson: But by setting them here won't the dependent targets have these values set anyway?

Makefile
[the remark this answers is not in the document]
jimmidyson: But that means introducing these files in this PR, and then removing them in a follow up - it feels better to me to make this change in this PR rather than a follow up, but I won't block on that.

controllers/nutanixcluster_controller.go
jimmidyson: 👍

controllers/nutanixcluster_controller.go
jimmidyson: Let's at least introduce the improved error check in this PR.

hack/install-go.sh
jimmidyson: 👍

Makefile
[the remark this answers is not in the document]
adiantum: No, it will affect precisely the target and not dependent.

controllers/nutanixcluster_controller.go
adiantum: Fixed

Makefile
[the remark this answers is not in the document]
adiantum: Fixed

Makefile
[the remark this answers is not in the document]
jimmidyson: I'm not sure I follow, this is being set as a global variable and as such will be used in every target that doesn't have a local override. Please correct me if I'm wrong 😅 

The condition `ifeq (${MAKECMDGOALS},docker-build-e2e)` means this `IMG_TAG` value is only being set when precisely `make docker-build-e2e` is run on the command line. If any other goals are specified then this will not be set. Is that actually what you want? By making these target specific variables you will get the same effect and they will be inherited regardless of whether other goals are specified or not, which provides more consistent usage.

Also, using the `MAKECMDGOALS` variable like this is pretty unusual and makes the Makefile harder for others to grok and imo should only be used when strictly necessary, which I'm not sure it is in this case.

Makefile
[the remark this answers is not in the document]
adiantum: @jimmidyson Sorry, some misunderstanding... The cluster-template.yaml manifest in release and manifest for E2E tests is not the same. Released manifest also requires patches. Find the way how to improve it is the case for future.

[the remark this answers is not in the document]
jimmidyson: So the released artifacts aren't usable directly? OK that is the context I'm missing! OK let's try to fix that up in future releases.

Makefile
dkoshkin, line 20: We've had good success with https://github.com/caarlos0/svu and then  `svu patch --pre-release dev` to get whatever the next dev version is.

I suspect this will get stale pretty quickly, but we can do a follow up PR.

templates/cluster-template.yaml
```diff
@@ -16,6 +234,35 @@ stringData:
...
+stringData:
+  nutanix-ccm-secret.yaml: "apiVersion: v1\nkind: Secret\nmetadata:\n  name: nutanix-creds\n
+    \ namespace: kube-system\nstringData:\n  credentials: |\n    [\n      {\n        \"type\":
+    \"basic_auth\", \n        \"data\": { \n          \"prismCentral\":{\n            \"username\":
+    \"${NUTANIX_USER}\",\n            \"password\": \"${NUTANIX_PASSWORD}\"\n          },\n
+    \         \"prismElements\": null\n        }\n      }\n    ]\n"
+type: addons.cluster.x-k8s.io/resource-set
+---
+apiVersion: addons.cluster.x-k8s.io/v1beta1
+kind: ClusterResourceSet
```
dkoshkin, line 250: Since this is the default now lets add this to `clusterctl.yaml`:
```
EXP_CLUSTER_RESOURCE_SET: "true"
```

api/v1beta1/nutanixmachine_types.go
```diff
@@ -115,6 +115,7 @@ type NutanixMachineStatus struct {
 	VmUUID string `json:"vmUUID,omitempty"`
 
 	// NodeRef is a reference to the corresponding workload cluster Node if it exists.
+	// Deprecated: Do not use. Will be removed in a future release.
```
dkoshkin, line 118: Can we remove the `NoreRef` column on line 142 since it will be empty?

Makefile
dkoshkin: Oh I see the same version needs to be used in test/e2e/config/nutanix.yaml

adiantum: @dkoshkin Good thing, but I think it should be done in separate PR with improvement of test scripts (including generation of e2e config). We already discuss it with @jimmidyson.

templates/cluster-template.yaml
adiantum: It's already there for E2E tests.

api/v1beta1/nutanixmachine_types.go
adiantum: @dkoshkin Good catch, thank you.

adiantum: Fixed

templates/cluster-template.yaml
dkoshkin: Thanks yes I see the e2e changes, but we should still change

dkoshkin: Thanks yes I see the e2e changes, but we should still change

adiantum: @dkoshkin Can we change it in separate PR as soon as it is developer tool, not for end users? Looks like there are bunch of dev tools improvement changes.

dkoshkin: 👍that works for me

dkoshkin (approved):
My comments were addressed, LGTM!

thunderboltsid (approved):

yannickstruyf3 (approved):

tuxtof (approved):
LGTM

Makefile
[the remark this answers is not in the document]
adiantum: @jimmidyson The behaviour that you describe with `docker-build-e2e` target is exactly what I expected.
The semantic of the `MAKECMDGOALS` var is very straightforward and imho pretty understandable even if you never used the variable.
We're going to make refactoring work, which I think should include the Makefile and dev scripts. Let's make improvements in separate PRs.

PR #346, opened 2023-12-13
build: gitingore .env and .envrc files

dkoshkin:
I use https://direnv.net/ for Go version management, having a local `.envrc` file makes it easier to add the correct exports and have it be automatically loaded.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,6 @@ github.com
 profile.cov
 coverage.xml
 
+# env files that usually contain secrets or local config
+.env
+.envrc


thunderboltsid (approved):

tuxtof (approved):

jimmidyson:
@dkoshkin For local values, it is better to add something like `dotenv_if_exists .envrc-local` or similar so that values in envrc can be specified or overridden locally, while still allowing for project wide envrc, as used by tooling like devbox.

PR #354, opened 2024-01-03
fix: improve error handling for Prism Client

faiq:
This adds some logic to properly format the prismCentral address string if a user provides an address with a scheme or port already included. It also adds error wrapping to clearly note where any issues might be coming from.

Fixes errors from bad user input 

previously if a user provided an address such as `https://prismcentral.com" this would cause a failure like this

```
2023-12-19T22:26:53Z    ERROR   Reconciler error        {"controller": "nutanixmachine", "controllerGroup": "infrastructure.cluster.x-k8s.io", "controllerKind": "NutanixMachine", "NutanixMachine": {"name":"faiq-test-mt-0-brp7d","namespace":"default"}, "namespace": "default", "name": "faiq-test-mt-0-brp7d", "reconcileID": "24ede42f-0ba6-41cd-a0b0-51c37c6f95d0", "error": "client auth error: Get \"https:///api/nutanix/v3/users/me\": http: no Host in request URL"}
``` 

this code now strips out the scheme and port number from the address if it is provided.

I deployed this to a local kind cluster. Previously I was seeing errors such as 

**Release note**:

```release-note

```


pkg/client/client_test.go
```diff
@@ -0,0 +1,52 @@
...
+	tests := []struct {
+		name            string
+		providedAddress string
+		expectedAddress string
+		expectedErr     error
+	}{
+		{
+			"with scheme",
+			"https://prism-bowser.ntnxsherlock.com",
+			"prism-bowser.ntnxsherlock.com",
```
deepakm-ntnx, line 18: pls add one test with ip address as well

pkg/client/client.go
```diff
@@ -63,15 +66,18 @@ func (n *NutanixClientHelper) GetClientFromEnvironment(ctx context.Context, nuta
...
+			return nil, err
 		}
+		prismCentralInfo.Address = address
 		if prismCentralInfo.Port == 0 {
 			return nil, fmt.Errorf("cannot get credentials if Prism Port is not set")
 		}
-		credentialRef := prismCentralInfo.CredentialRef
-		if credentialRef == nil {
-			return nil, fmt.Errorf("credentialRef must be set on prismCentral attribute for cluster %s in namespace %s", nutanixCluster.Name, nutanixCluster.Namespace)
+		credentialRef, err := GetCredentialRefForCluster(nutanixCluster)
```
deepakm-ntnx, line 77: line no 75: we can use ErrPrismAddressNotSet in return

pkg/client/client_test.go
faiq: added some tests to address this

pkg/client/client.go
faiq: I think the error here is more specific, but i'll move this error out to a var for readability

pkg/client/client.go
```diff
@@ -196,16 +202,48 @@ func GetCredentialRefForCluster(nutanixCluster *infrav1.NutanixCluster) (*creden
...
+		return u.Hostname(), nil
+	}
+	return address, nil
+}
+
+func parseIP(s string) (string, error) {
+	ip, _, err := net.SplitHostPort(s)
+	if err == nil {
+		return ip, nil
+	}
```
adiantum, line 241:
```suggestion
	ipStr, _, err := net.SplitHostPort(s)
	if err == nil {
		return ipStr, nil
	}
```

pkg/client/client.go
```diff
@@ -196,16 +202,48 @@ func GetCredentialRefForCluster(nutanixCluster *infrav1.NutanixCluster) (*creden
...
+
+	ip2 := net.ParseIP(s)
+	if ip2 == nil {
+		return "", fmt.Errorf("invalid IP %s", ip2)
+	}
+
+	return ip2.String(), nil
```
adiantum, line 248:
```suggestion
	ip := net.ParseIP(s)
	if ip == nil {
		return "", fmt.Errorf("invalid IP %s", ip)
	}

	return ip.String(), nil
```

adiantum (changes requested):
Let's improve var naming little bit.

thunderboltsid:
@faiq can you also rebase this against the latest changes in main.

pkg/client/client.go
```diff
@@ -196,16 +202,46 @@ func GetCredentialRefForCluster(nutanixCluster *infrav1.NutanixCluster) (*creden
...
-	if prismCentralinfo.CredentialRef.Kind != credentialTypes.SecretKind {
+	if prismCentralInfo.CredentialRef.Kind != credentialTypes.SecretKind {
 		return nil, nil
 	}
 
-	return prismCentralinfo.CredentialRef, nil
+	return prismCentralInfo.CredentialRef, nil
+}
+
+func validateAndSanitizePrismCentralInfoAddress(address string) (string, error) {
```
dkoshkin, line 219: @faiq what are your thoughts on doing all of this in validating/mutating webhooks? I think it would really clean up this file and simplify these functions.

faiq: there's been a lot of push back against doing this with validating webhooks.  #356 addresses the validation using CEL

dkoshkin: Even better 🙏

pkg/client/client.go
```diff
@@ -94,7 +98,7 @@ func (n *NutanixClientHelper) GetClientFromEnvironment(ctx context.Context, nuta
 	// Add env provider for CAPX manager
 	npe, err := n.getManagerNutanixPrismEndpoint()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to create prism endpoint %w", err)
```
jimmidyson, line 101: nit: Can we get `:` between new contextual error string and wrapped error?
```suggestion
		return nil, fmt.Errorf("failed to create prism endpoint: %w", err)
```

faiq: Done.

thunderboltsid (approved):

jimmidyson (approved):

tuxtof (approved):
LGTM

PR #355, opened 2024-01-05
refactor: task status file

dkoshkin:
This started as an effort to start increasing the code coverage, but noticed that there was some code we can remove and instead rely on the apimachine `wait` helper functions.

1. Added unit tests to cover existing functions
2. Removed the custom wait code (~notice how only the error message with the timeout changed, all other tests cases stayed the same~, all tests stayed the same after the refactor)
3. While working on the PR I kept getting mixed up on "state" vs "status" as they were being used interchangeably. I made it all consistent with "status" as that is what the API returns. 

Fixes #

New unit tests

I broke up the PR into logical commits and would recommend reviewing it that way too.

I initially started by trying to mock the client but the is a bug that prevented me from doing that https://github.com/nutanix-cloud-native/prism-go-client/pull/150. I went with the http route based on the client's own test files https://github.com/nutanix-cloud-native/prism-go-client/blob/main/v3/v3_service_test.go, but can refactor back to using mocks if we find that to be easier to maintain, if/when the bug PR is merged and released.

**Release note**:

```release-note

```

### DIFF
--- a/controllers/nutanixmachine_controller.go
+++ b/controllers/nutanixmachine_controller.go
@@ -666,7 +666,7 @@ func (r *NutanixMachineReconciler) getOrCreateVM(rctx *nctx.MachineContext) (*nu
 		return nil, errorMsg
 	}
 	log.Info(fmt.Sprintf("Waiting for task %s to get completed for VM %s", lastTaskUUID, rctx.NutanixMachine.Name))
-	err = nutanixClient.WaitForTaskCompletion(ctx, nc, lastTaskUUID)
+	err = nutanixClient.WaitForTaskToSucceed(ctx, nc, lastTaskUUID)
 	if err != nil {
 		errorMsg := fmt.Errorf("error occurred while waiting for task %s to start: %v", lastTaskUUID, err)
 		rctx.SetFailureStatus(capierrors.CreateMachineError, errorMsg)
--- a/pkg/client/status.go
+++ b/pkg/client/status.go
@@ -29,24 +29,24 @@ import (
 
 const (
 	pollingInterval = time.Second * 2
-	stateSucceeded  = "SUCCEEDED"
+	statusSucceeded = "SUCCEEDED"
 )
 
-// WaitForTaskCompletion will poll indefinitely every 2 seconds for the task with uuid to have status of "SUCCEEDED".
-// Returns an error from GetTaskState or a timeout error if the context is cancelled.
-func WaitForTaskCompletion(ctx context.Context, conn *nutanixClientV3.Client, uuid string) error {
+// WaitForTaskToSucceed will poll indefinitely every 2 seconds for the task with uuid to have status of "SUCCEEDED".
+// Returns an error from GetTaskStatus or a timeout error if the context is cancelled.
+func WaitForTaskToSucceed(ctx context.Context, conn *nutanixClientV3.Client, uuid string) error {
 	return wait.PollImmediateInfiniteWithContext(ctx, pollingInterval, func(ctx context.Context) (done bool, err error) {
-		state, getErr := GetTaskState(ctx, conn, uuid)
-		return state == stateSucceeded, getErr
+		status, getErr := GetTaskStatus(ctx, conn, uuid)
+		return status == statusSucceeded, getErr
 	})
 }
 
-func GetTaskState(ctx context.Context, client *nutanixClientV3.Client, taskUUID string) (string, error) {
+func GetTaskStatus(ctx context.Context, client *nutanixClientV3.Client, uuid string) (string, error) {
 	log := ctrl.LoggerFrom(ctx)
-	log.V(1).Info(fmt.Sprintf("Getting task with UUID %s", taskUUID))
-	v, err := client.V3.GetTask(ctx, taskUUID)
+	log.V(1).Info(fmt.Sprintf("Getting task with UUID %s", uuid))
+	v, err := client.V3.GetTask(ctx, uuid)
 	if err != nil {
-		log.Error(err, fmt.Sprintf("error occurred while waiting for task with UUID %s", taskUUID))
+		log.Error(err, fmt.Sprintf("error occurred while waiting for task with UUID %s", uuid))
 		return "", err
 	}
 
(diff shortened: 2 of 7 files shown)


pkg/client/status.go
```diff
@@ -0,0 +1,60 @@
...
+
+const (
+	pollingInterval = time.Second * 2
+	statusSucceeded = "SUCCEEDED"
+)
+
+// WaitForTaskToSucceed will poll indefinitely every 2 seconds for the task with uuid to have status of "SUCCEEDED".
+// Returns an error from GetTaskStatus or a timeout error if the context is cancelled.
+func WaitForTaskToSucceed(ctx context.Context, conn *nutanixClientV3.Client, uuid string) error {
+	return wait.PollImmediateInfiniteWithContext(ctx, pollingInterval, func(ctx context.Context) (done bool, err error) {
```
dkoshkin, line 38: The previous `Retry` was technically exponential, but it was only used in this package with initial and max intervals set to `2`, so really it was just polling every 2 seconds until an error or the context is cancelled.

pkg/client/status.go
```diff
@@ -0,0 +1,61 @@
...
+const (
+	pollingInterval = time.Second * 2
+	statusSucceeded = "SUCCEEDED"
+)
+
+// WaitForTaskToSucceed will poll indefinitely every 2 seconds for the task with uuid to have status of "SUCCEEDED".
+// The polling will not stop if the ctx is cancelled, it's only used for HTTP requests in the client.
+// WaitForTaskToSucceed will exit immediately on an error getting the task.
+func WaitForTaskToSucceed(ctx context.Context, conn *nutanixClientV3.Client, uuid string) error {
+	return wait.PollImmediateInfinite(pollingInterval, func() (done bool, err error) {
```
dkoshkin, line 39: The previous `Retry` was technically exponential, but it was only used in this package with initial and max intervals set to `2`, so really it was just polling every 2 seconds until an error ~or the context is cancelled~.
I initially misunderstood how the ctx was being used. It is **NOT** used to cancel the wait, but only passed into the client to its HTTP requests.

pkg/client/status_test.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2022 Nutanix
```
thunderboltsid, line 2: nit: make it 2024

test/helpers/prism-go-client/v3/client.go
```diff
@@ -1,3 +1,19 @@
+/*
+Copyright 2022 Nutanix
```
thunderboltsid, line 2: nit: make it 2024

controllers/nutanixmachine_controller.go
```diff
@@ -666,7 +666,7 @@ func (r *NutanixMachineReconciler) getOrCreateVM(rctx *nctx.MachineContext) (*nu
 		return nil, errorMsg
 	}
 	log.Info(fmt.Sprintf("Waiting for task %s to get completed for VM %s", lastTaskUUID, rctx.NutanixMachine.Name))
-	err = nutanixClient.WaitForTaskCompletion(ctx, nc, lastTaskUUID)
+	err = nutanixClient.WaitForTaskToSucceed(ctx, nc, lastTaskUUID)
```
thunderboltsid, line 669: nit: can we inline this?
```go
if err := ...; err != nil {
    ...
}
```

pkg/client/status.go
```diff
@@ -0,0 +1,61 @@
...
+		status, getErr := GetTaskStatus(ctx, conn, uuid)
+		return status == statusSucceeded, getErr
+	})
+}
+
+func GetTaskStatus(ctx context.Context, client *nutanixClientV3.Client, uuid string) (string, error) {
+	log := ctrl.LoggerFrom(ctx)
+	log.V(1).Info(fmt.Sprintf("Getting task with UUID %s", uuid))
```
adiantum, line 47: I remember that we would like to use structured log in future. Maybe refactoring PR is good place to start using structured logs?

controllers/nutanixmachine_controller.go
dkoshkin: The rest of the function is not inlining it, since this was existing code can we make it consistent through a linter instead?

test/helpers/prism-go-client/v3/client.go
```diff
@@ -0,0 +1,69 @@
...
+		return nil, fmt.Errorf("error creating Nutanix test client: %w", err)
+	}
+	return &TestClient{client, mux, server}, nil
+}
+
+func (c *TestClient) Close() {
+	c.server.Close()
+}
+
+func (c *TestClient) AddHandler(pattern string, handler func(w http.ResponseWriter, r *http.Request)) {
```
adiantum, line 63: Maybe AddMockHandler or something similar? This naming looks little bit confusing.

adiantum (commented):
Typo in PR name: "refacotr: ..."

test/helpers/prism-go-client/v3/client.go
dkoshkin: Good idea 👍

pkg/client/status.go
dkoshkin: I would like to do that in a different PR once I better understand what we are trying to achieve with structured logging and when we have some linters to help us out, if thats ok.
This change was more targeted removing code and adding unit tests, it didn't change any other "business logic".

controllers/nutanixmachine_controller.go
thunderboltsid: We'll have to do that in other PRs, then. But, sure, as long as we do get around to it.

pkg/client/status_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	nutanixtestclient "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/test/helpers/prism-go-client/v3"
+)
+
+func Test_GetTaskStatus(t *testing.T) {
+	client, err := nutanixtestclient.NewTestClient()
```
deepakm-ntnx, line 32: can we use "github.com/golang/mock/gomock" framework to mock interfaces and structs? eg: mockgen to create the wrappers for interfaces eg https://github.com/nutanix-core/k8s-agent/blob/main/k8s-agent/Makefile#L82 wherever possible

dkoshkin: I will refactor that once possible, but there is a bug in the SDK that prevents me from implementing the Service interface https://github.com/nutanix-cloud-native/prism-go-client/pull/150

thunderboltsid (approved):

adiantum (approved):
/lgtm
/approve
/meow

PR #357, opened 2024-01-09
Add `pkg_test.go` files for every package

thunderboltsid:
Unless a `pkg_test.go` file is present in a `pkg` package, go toolchain will not consider that package for the purposes of coverage metrics. This is a regression and will be fixed in go1.22 but until then we need these files in each package to keep the coverage report honest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -379,7 +379,7 @@ endif
 
 .PHONY: coverage
 coverage: setup-envtest ## Run the tests of the project and export the coverage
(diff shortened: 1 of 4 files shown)


yannickstruyf3 (approved):

adiantum (approved):
/lgtm
/approve
/woof

tuxtof (approved):
LGTM

PR #360, opened 2024-01-10
refactor: client.go file helper methods

dkoshkin:
Refactored the existing methods and functions to be unit testable.
Also made some methods that do not use the struct as generic functions.
The changes were primarily an effort to add unit test coverage.

Fixes #

New unit tests and ran through the dev doc to create a cluster with management creds.

**Release note**:

```release-note

```


api/v1beta1/nutanixcluster_types.go
```diff
@@ -86,12 +88,12 @@ type NutanixClusterStatus struct {
...
-//+kubebuilder:subresource:status
-//+kubebuilder:storageversion
-//+kubebuilder:printcolumn:name="ControlplaneEndpoint",type="string",JSONPath=".spec.controlPlaneEndpoint.host",description="ControlplaneEndpoint"
-//+kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.ready",description="in ready status"
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=nutanixclusters,shortName=ncl,scope=Namespaced,categories=cluster-api
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="ControlplaneEndpoint",type="string",JSONPath=".spec.controlPlaneEndpoint.host",description="ControlplaneEndpoint"
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.ready",description="in ready status"
```
dkoshkin, line 96: This was formatted automatically by the IDE.

api/v1beta1/nutanixcluster_types.go
```diff
@@ -145,7 +147,22 @@ func (ncl *NutanixCluster) SetConditions(conditions capiv1.Conditions) {
 	ncl.Status.Conditions = conditions
 }
 
-//+kubebuilder:object:root=true
+func (ncl *NutanixCluster) GetCredentialRefForCluster() (*credentialTypes.NutanixCredentialReference, error) {
```
dkoshkin, line 150: I think it makes sense for this to be a method of the struct.

controllers/helpers.go
```diff
@@ -47,12 +48,8 @@ const (
 func CreateNutanixClient(ctx context.Context, secretInformer coreinformers.SecretInformer, cmInformer coreinformers.ConfigMapInformer, nutanixCluster *infrav1.NutanixCluster) (*nutanixClientV3.Client, error) {
 	log := ctrl.LoggerFrom(ctx)
 	log.V(1).Info("creating nutanix client")
-	helper, err := nutanixClientHelper.NewNutanixClientHelper(secretInformer, cmInformer)
-	if err != nil {
-		log.Error(err, "error creating nutanix client helper")
-		return nil, err
-	}
-	return helper.GetClientFromEnvironment(ctx, nutanixCluster)
+	helper := nutanixClient.NewHelper(secretInformer, cmInformer)
```
dkoshkin, line 51: Renamed the function to `NewHelper` to avoid the helper stutter.

controllers/nutanixcluster_controller.go
```diff
@@ -101,11 +101,11 @@ func (r *NutanixClusterReconciler) SetupWithManager(ctx context.Context, mgr ctr
...
-//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;update;delete
-//+kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch
-//+kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixclusters,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixclusters/status,verbs=get;update;patch
-//+kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixclusters/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;update;delete
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixclusters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixclusters/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixclusters/finalizers,verbs=update
```
dkoshkin, line 108: Changed by the IDE

controllers/nutanixcluster_controller.go
```diff
@@ -418,3 +418,10 @@ func (r *NutanixClusterReconciler) reconcileCredentialRef(ctx context.Context, n
 	}
 	return nil
 }
+
+func mustGetCredentialRefForCluster(nutanixCluster *infrav1.NutanixCluster) (*credentialTypes.NutanixCredentialReference, error) {
```
dkoshkin, line 422: nutanixCluster being nil and returning an error seemed like the controller's concern, added a helper function here.

pkg/client/client.go
```diff
@@ -37,66 +36,139 @@ import (
...
+
+func (n *NutanixClientHelper) withCustomNutanixPrismEndpointReader(getter func() (*credentials.NutanixPrismEndpoint, error)) *NutanixClientHelper {
+	n.managerNutanixPrismEndpointReader = getter
+	return n
 }
 
-func (n *NutanixClientHelper) GetClientFromEnvironment(ctx context.Context, nutanixCluster *infrav1.NutanixCluster) (*nutanixClientV3.Client, error) {
+// BuildClientForNutanixClusterWithFallback builds a Nutanix Client from the information provided in nutanixCluster.
+func (n *NutanixClientHelper) BuildClientForNutanixClusterWithFallback(ctx context.Context, nutanixCluster *infrav1.NutanixCluster) (*nutanixClientV3.Client, error) {
+	me, err := n.buildManagementEndpoint(ctx, nutanixCluster)
```
dkoshkin, line 80: Pulled this logic out into its own function because testing `BuildClientForNutanixClusterWithFallback` is complicated since there is nothing really to compare `nutanixClientV3.Client` to as it has an interface and unexorted types in the struct.

pkg/client/client.go
```diff
@@ -37,66 +36,139 @@ import (
...
+	}
+}
+
+func (n *NutanixClientHelper) withCustomNutanixPrismEndpointReader(getter func() (*credentials.NutanixPrismEndpoint, error)) *NutanixClientHelper {
+	n.managerNutanixPrismEndpointReader = getter
+	return n
 }
 
-func (n *NutanixClientHelper) GetClientFromEnvironment(ctx context.Context, nutanixCluster *infrav1.NutanixCluster) (*nutanixClientV3.Client, error) {
+// BuildClientForNutanixClusterWithFallback builds a Nutanix Client from the information provided in nutanixCluster.
```
dkoshkin, line 78: Renamed the function to "Build..." because I thought it was more fitting what is happening, where a struct is being built from some information.

pkg/client/client.go
```diff
@@ -37,66 +36,139 @@ import (
...
-		additionalTrustBundleRef := prismCentralInfo.AdditionalTrustBundle
-		if additionalTrustBundleRef != nil &&
-			additionalTrustBundleRef.Kind == credentialTypes.NutanixTrustBundleKindConfigMap &&
-			additionalTrustBundleRef.Namespace == "" {
-			additionalTrustBundleRef.Namespace = nutanixCluster.Namespace
-		}
-		providers = append(providers, kubernetesEnv.NewProvider(
-			*nutanixCluster.Spec.PrismCentral,
-			n.secretInformer,
-			n.configMapInformer))
```
dkoshkin, line 93: I pulled this out into its own function to make the codepaths more testable, the logic did not change.

pkg/client/client.go
```diff
@@ -37,66 +36,139 @@ import (
...
+		additionalTrustBundleRef.Namespace == "" {
+		additionalTrustBundleRef.Namespace = nutanixCluster.Namespace
+	}
+
+	return kubernetesEnv.NewProvider(*prismCentralInfo, n.secretInformer, n.configMapInformer), nil
+}
+
+// buildProviderFromFile will return an envTypes.Provider with info from the provided file.
+// It will return an error if required information is missing.
+func (n *NutanixClientHelper) buildProviderFromFile() (envTypes.Provider, error) {
```
dkoshkin, line 170: Same thing here, pulled it out into its own function to make it more testable, but the logic stayed the same.

pkg/client/client.go
```diff
@@ -115,99 +187,64 @@ func (n *NutanixClientHelper) GetClientFromEnvironment(ctx context.Context, nuta
...
 	}
-
-	return n.GetClient(creds, me.AdditionalTrustBundle)
+	return cli, nil
 }
 
-func (n *NutanixClientHelper) GetClient(cred prismgoclient.Credentials, additionalTrustBundle string) (*nutanixClientV3.Client, error) {
-	if cred.Username == "" {
-		return nil, fmt.Errorf("could not create client because username was not set")
+func buildClientFromCredentials(creds prismgoclient.Credentials, additionalTrustBundle string) (*nutanixClientV3.Client, error) {
```
dkoshkin, line 207: Split this out to be unit testable, it would have been more difficult to get the `cli.V3.GetCurrentLoggedInUser(context.Background())` call working with some fake http server.

pkg/client/client.go
```diff
@@ -115,99 +187,64 @@ func (n *NutanixClientHelper) GetClientFromEnvironment(ctx context.Context, nuta
...
+	if creds.Username == "" {
```
dkoshkin, line 208: Renamed to `creds` for consistency with another function here.

pkg/client/client.go
```diff
@@ -115,99 +187,64 @@ func (n *NutanixClientHelper) GetClientFromEnvironment(ctx context.Context, nuta
...
-
-func (n *NutanixClientHelper) readEndpointConfig() ([]byte, error) {
-	if b, err := os.ReadFile(filepath.Join(configPath, endpointKey)); err == nil {
-		return b, err
-	} else if os.IsNotExist(err) {
-		return []byte{}, nil
-	} else {
-		return []byte{}, err
-	}
-}
```
dkoshkin, line 195: Removed this and just replaced with `os.ReadFile` directly, this method was **not** failing with an empty file, but then `if npe.CredentialRef == nil {` would fail with an misleading error since the struct is just empty.
Instead this will now fail if the file is not found.

pkg/client/client.go
```diff
@@ -37,66 +36,139 @@ import (
...
 
-func NewNutanixClientHelper(secretInformer coreinformers.SecretInformer, cmInformer coreinformers.ConfigMapInformer) (*NutanixClientHelper, error) {
+func NewHelper(secretInformer coreinformers.SecretInformer, cmInformer coreinformers.ConfigMapInformer) *NutanixClientHelper {
 	return &NutanixClientHelper{
-		secretInformer:    secretInformer,
-		configMapInformer: cmInformer,
-	}, nil
+		secretInformer:                    secretInformer,
+		configMapInformer:                 cmInformer,
+		managerNutanixPrismEndpointReader: readManagerNutanixPrismEndpointFromDefaultFile,
```
faiq, line 69: this should also be something we should be able to pass into the constructor

dkoshkin: Previously this package held the constant `configPath` and knew how to read that file.
I agree, thats probably more of a concern of the controller, I'll pull this out as you suggested, thanks!

dkoshkin: Actually after making these changes I don't think it makes much sense to take this in in the construction either.
At some point we should refactor this more and have the API similar to other providers' `identityRef`, e.g. https://cluster-api-aws.sigs.k8s.io/topics/multitenancy.

What do you think keeping it the way it is now with a private decorator thats used in the unit tests, and then working on a further API changes in the future?

dkoshkin:
Thanks for triggering the tests @thunderboltsid, do you think these test failures are related, from the errors it looks like it couldn't delete the old Nodes, is that a known flake?

thunderboltsid:
> Thanks for triggering the tests @thunderboltsid, do you think these test failures are related, from the errors it looks like it couldn't delete the old Nodes, is that a known flake?

The issues stemmed from some older resource from another failed test that were not cleaned up properly that ended up causing IP contention between two separate clusters.

thunderboltsid:
Let's rebase this PR against the latest changes; I've manually cleaned up the problematic resources so it shouldn't be a problem now.

api/v1beta1/nutanixcluster_types_test.go
```diff
@@ -0,0 +1,114 @@
...
+					PrismCentral: &credentials.NutanixPrismEndpoint{
+						CredentialRef: &credentials.NutanixCredentialReference{
+							Kind: "unknown",
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "prismCentralInfo isn not nil but CredentialRef is nil, should fail",
```
thunderboltsid, line 90: typo: `isn`

controllers/nutanixcluster_controller.go
thunderboltsid: Not sure if it's just me but when I see `must` in a function name, I expect that function to panic instead of error. Given it's being used in just two places, we can just do a nil check in those places. Alternatively, we can consider renaming this function to something like `GetNutanixClusterCredentialRef`.

api/v1beta1/nutanixcluster_types.go
thunderboltsid: Consider streamlining the name while at it: Something like `PrismCredentialRef` or `GetPrismCredentialRef` stutters less.

pkg/client/client.go
thunderboltsid: @dkoshkin fine by me.

pkg/client/client_test.go
```diff
@@ -1,12 +1,761 @@
...
+hRSGlVuXMlBvPci6Zgzj/L24ScF2iUkZ/cCovYmjZy/Gn7xxGWC4LeksyZB2ZnuU4q941mVTXTzW
+nLLPKQP5L6RQstRIzgUyVYr9smRMDuSYB3Xbf9+5CFVghTAp+XtIpGmG4zU/HoZdenoVve8AjhUi
+VBcAkCaTvA5JaJG/+EfTnZVCwQ5N328mz8MYIWJmQ3DW1cAH4QIDAQABo0IwQDAOBgNVHQ8BAf8E
+BAMCAQYwDwYDVR0TAQH/BAUwAwEB/zAdBgNVHQ4EFgQUVeSB0RGAvtiJuQijMfmhJAkWuXAwDQYJ
+KoZIhvcNAQEFBQADggEBADubj1abMOdTmXx6eadNl9cZlZD7Bh/KM3xGY4+WZiT6QBshJ8rmcnPy
+T/4xmf3IDExoU8aAghOY+rat2l098c5u9hURlIIM7j+VrxGrD9cv3h8Dj1csHsm7mhpElesYT6Yf
+zX1XEC+bBAlahLVu2B064dae0Wx5XnkcFMXj0EyTO2U87d89vqbllRrDtRnDvV5bu/8j72gZyxKT
+J1wDLW8w0B62GqzeWvfRqqgnpv55gcR5mTNXuhKwqeBCbJPKVt7+bYQLCIt+jerXmCHG8+c8eS9e
+nNFMFY3h7CI3zJpDC5fcgJCNs2ebb0gIFVbPv/ErfF6adulZkMV8gzURZVE=
+-----END CERTIFICATE-----`
```
thunderboltsid, line 105: nit: can we store these in files and embed them here?

pkg/client/client_test.go
```diff
@@ -1,12 +1,761 @@
...
+	  "credentialRef": {
+        "kind": "Secret",
+        "name": "creds",
+        "namespace": "test"
+	  }
+	}`
+	invalidTestConfigMissingCredentialRef = `{
+      "address": "cluster-endpoint",
+      "port": 9440
+	}`
```
thunderboltsid, line 119: nit: can we store the multi-line content in files and embed them here?

pkg/client/client_test.go
```diff
@@ -1,12 +1,761 @@
...
+				assert.ErrorContains(t, err, tt.expectedErrString)
+			}
+			assert.Equal(t, tt.expectedEndpoint, endpoint)
+		})
+	}
+}
+
+func Test_readManagerNutanixPrismEndpointFromFile_IsNotExist(t *testing.T) {
+	_, err := readManagerNutanixPrismEndpointFromFile("filedoesnotexist.json")
+	assert.ErrorIs(t, err, os.ErrNotExist, err)
```
thunderboltsid, line 729: why is `err` referred here twice?

pkg/client/client_test.go
```diff
@@ -1,12 +1,761 @@
...
+			config:            []byte(invalidTestConfigMissingCredentialRef),
+			expectedErrString: "credentialRef must be set on CAPX manager",
+		},
+	}
+	for _, tt := range tests {
+		tt := tt // Capture range variable.
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			f, err := os.CreateTemp(t.TempDir(), "config-*.json")
+			assert.NoError(t, err, "error creating temp file")
```
thunderboltsid, line 714: this should be a `require.NoError`.

pkg/client/client_test.go
```diff
@@ -1,12 +1,761 @@
...
+			_, err = f.Write(tt.config)
+			assert.NoError(t, err, "error writing temp file")
```
thunderboltsid, line 716: this should be a `require.NoError`.

pkg/client/client.go
```diff
@@ -37,66 +36,139 @@ import (
...
-	configPath          = "/etc/nutanix/config"
-	endpointKey         = "prismCentral"
-	capxNamespaceKey    = "POD_NAMESPACE"
+
+	capxNamespaceKey = "POD_NAMESPACE"
+)
+
+var (
+	// use a var to override in tests
+	configPath = "/etc/nutanix/config/prismCentral"
```
thunderboltsid, line 45: Given we are using `readManagerNutanixPrismEndpointFromFile` in tests, can we can make this a const?

pkg/client/client_test.go
dkoshkin: Oops, no need for it, the error will already be clear what the error we got is vs what the expected one

pkg/client/client_test.go
dkoshkin: Yep good catch, test can't proceed if that fails

pkg/client/client.go
dkoshkin: Yep forgot in the refactor, thanks!

pkg/client/client_test.go
dkoshkin: 👍 that makes sense to me

api/v1beta1/nutanixcluster_types.go
dkoshkin: 🤦‍♂️ forgot to actually do what I left a comment on, fixed in https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/360/commits/d8b6f454525affd1c7cd5347182fc2614d4207a5

thunderboltsid (approved):

thunderboltsid (approved):

faiq (approved):

PR #361, opened 2024-01-15
Update codecov.yml to omit generated code

thunderboltsid:
Everytime we add or modify a CRD, we will fail to meet generated code coverage. The onus of ensuring the correctness of generated code lies on the generator itself. We can, in future, think about generating tests for this code (trust but verify) but, for the time being, it's best to exclude this from metrics.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,6 +1,10 @@
 codecov:
   require_ci_to_pass: yes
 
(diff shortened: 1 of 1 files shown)


tuxtof (approved):

adiantum (approved):
/lgtm
/approve

PR #364, opened 2024-01-17
Update Labels for E2E tests

thunderboltsid:
- Replace 'prblocker' with 'quickstart' to clarify intent
- Remove 'slow' and 'network' as every test had it making it moot
- Expand scope of 'capx-feature-test' to target all nutanix interactions
- Add 'scaling' to target all tests involving node churn


tuxtof (approved):
LGTM

PR #367, opened 2024-01-18
Check-in v1.3.0 manifests in test/e2e/data

thunderboltsid:
v1.3.0 manifests are to to spin up a point in time infrastructure from
a particular release. We also updated the versions to ensure there's a
version that e2e tests can target. We have also added a
'release-blocker' comment everywhere that needs to be modified to ensure
a release does not break e2es.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,8 @@ IMG_TAG := latest
 endif
 
 ifeq (${LOCAL_PROVIDER_VERSION},latest)
-# Change this versions after release when required here and in e2e config (test/e2e/config/nutanix.yaml)
-LOCAL_PROVIDER_VERSION := v1.3.99
+# TODO(release-blocker): Change this versions after release when required here and in e2e config (test/e2e/config/nutanix.yaml)
+LOCAL_PROVIDER_VERSION := v1.4.99
 endif
 
 # PLATFORMS is a list of platforms to build for.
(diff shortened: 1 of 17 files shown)


deepakm-ntnx:
we need to update test code as well. pls see ref to 1.2.4 https://github.com/search?q=repo%3Anutanix-cloud-native%2Fcluster-api-provider-nutanix%20v1.2.4&type=code

thunderboltsid:
> https://github.com/search?q=repo%3Anutanix-cloud-native%2Fcluster-api-provider-nutanix%20v1.2.4&type=code
Good catch, does it need to be in the same change?

deepakm-ntnx:
> > https://github.com/search?q=repo%3Anutanix-cloud-native%2Fcluster-api-provider-nutanix%20v1.2.4&type=code
> 
> Good catch, does it need to be in the same change?

Preferred so that we can make this as a ref PR listing all the changes we need to make during every release?

adiantum (approved):
/lgtm
/approve

adiantum:
> we need to update test code as well. pls see ref to 1.2.4 https://github.com/search?q=repo%3Anutanix-cloud-native%2Fcluster-api-provider-nutanix%20v1.2.4&type=code

Actually we don't need to do it.
We can add additional test with additional version if required

faiq (approved):

Makefile
```diff
@@ -304,11 +304,14 @@ undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/confi
 ##@ Templates
 
 .PHONY: cluster-e2e-templates
-cluster-e2e-templates: $(KUSTOMIZE) cluster-e2e-templates-v1beta1 cluster-e2e-templates-v1alpha4 cluster-e2e-templates-v124 ## Generate cluster templates for all versions
+cluster-e2e-templates: $(KUSTOMIZE) cluster-e2e-templates-v1beta1 cluster-e2e-templates-v1alpha4 cluster-e2e-templates-v124 cluster-e2e-templates-v130 ## Generate cluster templates for all versions
```
deepakm-ntnx, line 307: lets tag release blocker at all places then so that we know all the places which needs change

deepakm-ntnx: better would be to document in workflow doc

PR #378, opened 2024-02-09
update kube-vip to 0.7.0

tuxtof:

update kube-vip to v0.7.0

**Release note**:

```release-note
update kube-vip to v0.7.0
```

### DIFF
--- a/test/e2e/data/infrastructure-nutanix/v1.2.4/base/cluster-with-kcp.yaml
+++ b/test/e2e/data/infrastructure-nutanix/v1.2.4/base/cluster-with-kcp.yaml
@@ -81,7 +81,7 @@ spec:
           spec:
             containers:
               - name: kube-vip
-                image: ghcr.io/kube-vip/kube-vip:v0.6.4
+                image: ghcr.io/kube-vip/kube-vip:v0.7.0
                 imagePullPolicy: IfNotPresent
                 args:
                   - manager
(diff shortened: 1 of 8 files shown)


yannickstruyf3 (approved):

test/e2e/data/infrastructure-nutanix/v1.3.0/base/cluster-with-kcp.yaml
```diff
@@ -81,7 +81,7 @@ spec:
           spec:
             containers:
               - name: kube-vip
-                image: ghcr.io/kube-vip/kube-vip:v0.6.4
+                image: ghcr.io/kube-vip/kube-vip:v0.7.0
```
thunderboltsid, line 84: IMO, we should leave the versioned templates as is so they reflect the status of the template as of that release.

test/e2e/data/infrastructure-nutanix/v1.2.4/base/cluster-with-kcp.yaml
```diff
@@ -81,7 +81,7 @@ spec:
           spec:
             containers:
               - name: kube-vip
-                image: ghcr.io/kube-vip/kube-vip:v0.6.4
+                image: ghcr.io/kube-vip/kube-vip:v0.7.0
```
thunderboltsid, line 84: IMO, we should leave the versioned templates as is so they reflect the status of the template as of that release.

PR #388, opened 2024-03-12
Mark v1alpha4 CRDs as Deprecated and Unserved

thunderboltsid:
CAPI v1.6 has v1alpha4 types marked as deprecated and unserved. To stay in sync, we are doing the same. With v1.7 these will be removed entirely and we will follow suit when we update to v1.7.

### DIFF
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixclusters.yaml
@@ -27,6 +27,9 @@ spec:
       jsonPath: .status.ready
       name: Ready
       type: string
+    deprecated: true
+    deprecationWarning: v1alpha4 NutanixCluster is deprecated and marked as unserved.
+      It will be removed in a future release. Use v1beta1 NutanixCluster instead.
     name: v1alpha4
     schema:
       openAPIV3Schema:
@@ -300,7 +303,7 @@ spec:
                 type: boolean
             type: object
         type: object
-    served: true
+    served: false
     storage: false
     subresources:
       status: {}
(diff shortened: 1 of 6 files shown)


config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixclusters.yaml
```diff
@@ -27,6 +27,9 @@ spec:
       jsonPath: .status.ready
       name: Ready
       type: string
+    deprecated: true
+    deprecationWarning: v1alpha4 is deprecated and set as unserved and  will be removed
```
deepakm-ntnx, line 31: nit: extra space between "and" and "will" in the stmt

config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixclusters.yaml
```diff
@@ -300,7 +303,7 @@ spec:
                 type: boolean
             type: object
         type: object
-    served: true
+    served: false
```
deepakm-ntnx, line 306: can u pls point to the docs for the the meaning of served for my understanding

thunderboltsid: https://book.kubebuilder.io/reference/markers/crd.html

The `+kubebuilder:unserved` annotation in Kubebuilder is used to indicate that a particular version of a Custom Resource Definition (CRD) should not be served by the Kubernetes API server.

The Kubernetes API server will not allow creation or modification of resources using the unserved version. However, it can still manage resources that were created with that version if they are stored in etcd. This is useful for ensuring that no new resources are created with the deprecated version while still maintaining the ability to work with existing resources during a migration period.

config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixclusters.yaml
thunderboltsid: good catch. Updated to remove the space and updated the comment to be more descriptive.

tuxtof (approved):
LGTM

PR #393, opened 2024-03-15
Add variable for setting failureDomains on Cluster

thunderboltsid:
Use conditional presence of failureDomain to apply two patches
- one to remove cluster and subnet for controlPlane Nutanix machine template
- one to add failureDomains from variable

The PR also adds a template-test make target which triggers a test that is used to verify that clusterclass patches have the behaviour we expect. This spins up a kind cluster but is not a e2e test as no real underlying infrastructure resources are created. This is not a unit test either as it doesn't exercise any actual code in CAPX. The PR also adds template-test to the GitHub action to integrate it in CI toolchain.


deepakm-ntnx (approved):

templates/template_test.go
```diff
@@ -0,0 +1,279 @@
...
+		_ = teardownTestEnvironment()
+	}()
+
+	tt := []struct {
+		name            string
+		clusterManifest string
+		assertions      func(t *testing.T, clnt client.Client, clusterName string)
+	}{
+		{
+			name:            "TestClusterClassPatchesWithFailureDomain",
```
deepakm-ntnx, line 244: could you please add some comments of the purpose of each test for readability

Makefile
```diff
@@ -307,7 +307,7 @@ mocks: ## Generate mocks for the project
 	mockgen -destination=mocks/k8sclient/cm_informer.go -package=mockk8sclient k8s.io/client-go/informers/core/v1 ConfigMapInformer
 	mockgen -destination=mocks/k8sclient/secret_informer.go -package=mockk8sclient k8s.io/client-go/informers/core/v1 SecretInformer
 
-GOTESTPKGS = $(shell go list ./... | grep -v /mocks)
+GOTESTPKGS = $(shell go list ./... | grep -v /mocks | grep -v /templates)
```
deepakm-ntnx, line 310: do we need tee instead of pipe for 2 greps to work or is templates inside /mocks

templates/testdata/cluster-with-failure-domain.yaml
```diff
@@ -0,0 +1,79 @@
...
+              name: cluster1
+              type: name
+            controlPlane: true
+            name: cluster1
+            subnets:
+              - name: fake-subnet
+                type: name
+          - cluster:
+              name: cluster2
+              type: name
```
deepakm-ntnx, line 60: we should add a test with uuid cluster as well. same for subnet

thunderboltsid: will update the existing test to have one cluster and subnet defined w/ UUIDs

templates/ccm/nutanix-ccm-cm.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
deepakm-ntnx, line 1: in base pr this is no more present.

templates/topology/nutanix-ccm-cm.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
deepakm-ntnx, line 1: this is also removed in base pr

templates/ccm/nutanix-ccm-cm.yaml
thunderboltsid: Good catch

templates/topology/nutanix-ccm-cm.yaml
thunderboltsid: Good catch

deepakm-ntnx (approved):

dkoshkin (approved):
Thanks for the new tests

PR #394, opened 2024-03-19
Add cluster class patch for kubevip control plane endpoint 

thunderboltsid:
Prepend sed command on KCPT prekubadm commands list. The sed commands mutate the kube-vip template with variables to configure kube-vip.


templates/clusterclass/kcpt.yaml
```diff
@@ -40,9 +40,9 @@ spec:
                       - name: vip_arp
                         value: "true"
                       - name: address
-                        value: "${CONTROL_PLANE_ENDPOINT_IP}"
+                        value: "control_plane_endpoint_ip"
                       - name: port
-                        value: "${CONTROL_PLANE_ENDPOINT_PORT=6443}"
+                        value: "control_plane_endpoint_port"
```
deepakm-ntnx, line 45: nit: pls file a tracker to add them as vars or default the values on line 63, 65 and 67 below

deepakm-ntnx: Also as discussed in future we may have to make the entire section as variable if user want to update any other variables which we are not sed-ing yet

deepakm-ntnx (dismissed):

deepakm-ntnx (approved):

templates/clusterclass/kcpt.yaml
deepakm-ntnx: as discussed earlier, we will add later in separate PR if needed

templates/cluster-template-clusterclass.yaml
```diff
@@ -108,6 +108,22 @@ spec:
         kind: NutanixClusterTemplate
         matchResources:
           infrastructureCluster: true
+    - jsonPatches:
```
dkoshkin, line 111: There are other envs in the template `KUBEVIP_SVC_ENABLE`, `KUBEVIP_LB_ENABLE` , etc

What do you think about adding a patch with the whole content of this file?
This way we can still have direct variable substation and allow users to override the whole thing if they need to change other parts of it. 

It also has the benefit of keeping the templates clean and reusable in other usecases

dkoshkin (changes requested):

templates/cluster-template-clusterclass.yaml
faiq: I agree. Also I don't understand why its being done with sed in a preKubeadmCommnd

deepakm-ntnx: already tracked thru https://jira.nutanix.com/browse/KRBN-8114 regarding extra parameter related changes

dkoshkin (approved):
Approving these changes, but lets please come back it for https://jira.nutanix.com/browse/KRBN-8114

PR #395, opened 2024-03-20
Add variable for setting the project on cluster resource

thunderboltsid:
The change also includes a patch to set the project property on nutanixmachinetemplate resources for control plane and worker machine deployments.

### DIFF
--- a/templates/clusterclass/clusterclass.yaml
+++ b/templates/clusterclass/clusterclass.yaml
@@ -263,6 +263,31 @@ spec:
             path: /spec/template/spec/cluster
           - op: remove
             path: /spec/template/spec/subnet
+  - name: add-project
+    enabledIf: "{{if .project}}true{{end}}"
+    definitions:
+      - selector:
+          apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+          kind: NutanixMachineTemplate
+          matchResources:
+            controlPlane: true
+        jsonPatches:
+          - op: add
+            path: /spec/template/spec/project
+            valueFrom:
+              variable: project
+      - selector:
+          apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+          kind: NutanixMachineTemplate
+          matchResources:
+            machineDeploymentClass:
+              names:
+                - nutanix-quick-start-worker
+        jsonPatches:
+          - op: add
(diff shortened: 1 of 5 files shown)


templates/clusterclass/clusterclass.yaml
```diff
@@ -263,6 +263,35 @@ spec:
...
+          matchResources:
+            machineDeploymentClass:
+              names:
+                - nutanix-quick-start-worker
+        jsonPatches:
+          - op: add
+            path: /spec/template/spec/project
+            valueFrom:
+              template: |
+                type: name
```
deepakm-ntnx, line 293: we should support uuid as well by adding type as part of variable itself

templates/clusterclass/clusterclass.yaml
```diff
@@ -376,3 +405,9 @@ spec:
                     enum:
                       - name
                       - uuid
+  - name: projectName
```
deepakm-ntnx, line 408: should be project and allow specifying name and uuid both types

templates/testdata/cluster-with-project.yaml
```diff
@@ -0,0 +1,54 @@
...
+          bootType: legacy
+          clusterName: fake-cluster
+          imageName: ubuntu-2204-kube-v1.29.2.qcow2
+          memorySize: 4Gi
+          subnetName: fake-subnet
+          systemDiskSize: 40Gi
+          vcpuSockets: 2
+          vcpusPerSocket: 1
+      - name: projectName
+        value: fake-project
```
deepakm-ntnx, line 48: pls add a test for project with uuid as well after supporting uuid as a type

templates/clusterclass/clusterclass.yaml
deepakm-ntnx: we can take uuid in all the vars wherever application in separate PR

deepakm-ntnx (approved):
lets add uuid support in separate PR if needed

deepakm-ntnx (approved):

thunderboltsid:
> lets add uuid support in separate PR if needed

I have updated the variable to admin `project` instead of `projectName` and it now admits a type, name, and uuid instead of just a name. PTAL.

deepakm-ntnx (approved):

PR #398, opened 2024-03-22
Switch Nutanix Client to using Session Auth

thunderboltsid:
This will ensure we make fewer basic auth requests to Prism Central IAM Services.


deepakm-ntnx:
it makes sense to have sessionAuth set when we create the first connection by getting the cookies thru /users/me call here https://github.com/nutanix-cloud-native/prism-go-client/blob/main/internal/client.go#L276

deepakm-ntnx:
we need to add a test case somehow to test this as well

deepakm-ntnx (approved):
lets add a test case in this or thru other PR

thunderboltsid:
> lets add a test case in this or thru other PR

Can you elaborate what you expect out of a test case?

deepakm-ntnx:
> > lets add a test case in this or thru other PR
> 
> Can you elaborate what you expect out of a test case?

we can start with following tests:
- create connection with sessionAuth set and observe that cookie is set => ideally this can be part of prism-go-client
- create 100 connections/clusters with sessionAuth set vs not set and see if there is any impact as scale was the main reason we are doing this
- etc

yannickstruyf3 (approved):

deepakm-ntnx:
> /hold

@thunderboltsid can we get this patch in and then add tests separately? we need to start the testing with scale test setup as well.

thunderboltsid:
> > /hold
> 
> @thunderboltsid can we get this patch in and then add tests separately? we need to start the testing with scale test setup as well.

any testing can be done using an image built from the branch. Merging this change alone won't fix anything.

pkg/client/client.go
```diff
@@ -191,16 +192,7 @@ func (n *NutanixClientHelper) buildProviderFromFile() (envTypes.Provider, error)
...
-	cli, err := buildClientFromCredentials(creds, additionalTrustBundle)
-	if err != nil {
-		return nil, err
-	}
-	// Check if the client is working
-	_, err = cli.V3.GetCurrentLoggedInUser(context.Background())
-	if err != nil {
-		return nil, fmt.Errorf("failed to get current logged in user with client: %w", err)
-	}
-	return cli, nil
```
thunderboltsid, line 203: With session-auth, this happens automatically as basic auth credentials are used to generate a session cookie by making a GetCurrentLoggedInUser call internally.

thunderboltsid:
Tests seem to be failing due to a change that broke the tests on main. We are reverting the change in https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/414

controllers/nutanixcluster_controller.go
```diff
@@ -170,20 +170,11 @@ func (r *NutanixClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		log.V(1).Info(fmt.Sprintf("Patched NutanixCluster. Status: %+v", cluster.Status))
 	}()
 
-	err = r.reconcileCredentialRef(ctx, cluster)
```
dkoshkin, line 173: What do you think about keeping `r.reconcileCredentialRef` but also returning the `Secret` and then passing that Secret into `getPrismCentralClientForCluster`, that would make the function a little more focused and more unit testable?

It would also make the changes in this PR smaller

thunderboltsid: I've narrowed down the focus of this PR to just enabling session-auth. I've moved the client caching scope to another PR. https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/415

dkoshkin (approved):

adiantum (approved):
/lgtm
/approve

PR #399, opened 2024-03-22
Add a cache for storing NutanixClients

thunderboltsid:
The cache stores a prismgoclient.V3 client instance for each NutanixCluster instance. The cache is shared between nutanixcluster and nutanixmachine controllers.

### DIFF
--- a/controllers/nutanixcluster_controller.go
+++ b/controllers/nutanixcluster_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	stderrors "errors"
 	"fmt"
 	"time"
 
@@ -44,6 +45,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	infrav1 "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/api/v1beta1"
+	"github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkg/client/prismclientcache"
 	nctx "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkg/context"
 )
 
@@ -178,10 +180,21 @@ func (r *NutanixClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 	conditions.MarkTrue(cluster, infrav1.CredentialRefSecretOwnerSetCondition)
 
-	v3Client, err := CreateNutanixClient(ctx, r.SecretInformer, r.ConfigMapInformer, cluster)
+	v3Client, err := prismclientcache.DefaultCache.Get(cluster.Name)
 	if err != nil {
-		conditions.MarkFalse(cluster, infrav1.PrismCentralClientCondition, infrav1.PrismCentralClientInitializationFailed, capiv1.ConditionSeverityError, err.Error())
-		return ctrl.Result{Requeue: true}, fmt.Errorf("nutanix client error: %v", err)
+		if stderrors.Is(err, prismclientcache.ErrorClientNotFound) {
+			log.Info("Nutanix Prism client not found in cache; Creating new client")
+			v3Client, err = CreateNutanixClient(ctx, r.SecretInformer, r.ConfigMapInformer, cluster)
+			if err != nil {
+				conditions.MarkFalse(cluster, infrav1.PrismCentralClientCondition, infrav1.PrismCentralClientInitializationFailed, capiv1.ConditionSeverityError, err.Error())
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, fmt.Errorf("nutanix prism client error: %w", err)
+			}
+			prismclientcache.DefaultCache.Set(cluster.Name, v3Client)
+			conditions.MarkTrue(cluster, infrav1.PrismCentralClientCondition)
+		} else {
+			conditions.MarkFalse(cluster, infrav1.PrismCentralClientCondition, infrav1.PrismCentralClientInitializationFailed, capiv1.ConditionSeverityError, err.Error())
+			return ctrl.Result{RequeueAfter: 5 * time.Second}, fmt.Errorf("nutanix prism client error: %w", err)
+		}
 	}
 	conditions.MarkTrue(cluster, infrav1.PrismCentralClientCondition)
 
@@ -239,6 +252,8 @@ func (r *NutanixClusterReconciler) reconcileDelete(rctx *nctx.ClusterContext) (r
 		Name:      rctx.Cluster.Name,
 	}
 	nctx.RemoveRemoteClient(clusterKey)
+	// delete the client from the cache
+	prismclientcache.DefaultCache.Delete(rctx.NutanixCluster.Name)
 
 	return reconcile.Result{}, nil
 }
--- a/controllers/nutanixmachine_controller.go
+++ b/controllers/nutanixmachine_controller.go
@@ -50,6 +50,7 @@ import (
 
 	infrav1 "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/api/v1beta1"
 	nutanixClient "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkg/client"
+	"github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkg/client/prismclientcache"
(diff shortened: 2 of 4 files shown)


controllers/nutanixcluster_controller.go
```diff
@@ -362,7 +375,7 @@ func (r *NutanixClusterReconciler) reconcileCredentialRefDelete(ctx context.Cont
 
 	if secret.DeletionTimestamp.IsZero() {
 		log.Info(fmt.Sprintf("removing secret %s in namespace %s for cluster %s", secret.Name, secret.Namespace, nutanixCluster.Name))
-		if err := r.Client.Delete(ctx, secret); err != nil && !errors.IsNotFound(err) {
+		if err := r.Client.Delete(ctx, secret); err != nil && !k8sApiErrors.IsNotFound(err) {
```
deepakm-ntnx, line 378: we need to remove it from cache too here

controllers/nutanixmachine_controller.go
```diff
@@ -232,12 +233,22 @@ func (r *NutanixMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{Requeue: true}, nil
 	}
 
-	v3Client, err := CreateNutanixClient(ctx, r.SecretInformer, r.ConfigMapInformer, ntxCluster)
+	v3Client, err := prismclientcache.DefaultCache.Get(ntxCluster.Name)
 	if err != nil {
-		conditions.MarkFalse(ntxMachine, infrav1.PrismCentralClientCondition, infrav1.PrismCentralClientInitializationFailed, capiv1.ConditionSeverityError, err.Error())
-		return ctrl.Result{Requeue: true}, fmt.Errorf("client auth error: %v", err)
+		if errors.Is(err, prismclientcache.ErrorClientNotFound) {
+			log.Info("Nutanix client not found in cache. Creating new client")
```
deepakm-ntnx, line 239: nit: Nutanix client not found in cache. => Nutanix pc client is not found in cache.

controllers/nutanixcluster_controller.go
```diff
@@ -178,12 +182,21 @@ func (r *NutanixClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 	conditions.MarkTrue(cluster, infrav1.CredentialRefSecretOwnerSetCondition)
 
-	v3Client, err := CreateNutanixClient(ctx, r.SecretInformer, r.ConfigMapInformer, cluster)
+	v3Client, err := prismclientcache.DefaultCache.Get(cluster.Name)
 	if err != nil {
-		conditions.MarkFalse(cluster, infrav1.PrismCentralClientCondition, infrav1.PrismCentralClientInitializationFailed, capiv1.ConditionSeverityError, err.Error())
-		return ctrl.Result{Requeue: true}, fmt.Errorf("nutanix client error: %v", err)
+		if errors.Is(err, prismclientcache.ErrorClientNotFound) {
+			log.Info("Nutanix client not found in cache. Creating new client")
```
deepakm-ntnx, line 188: nit: Nutanix client not found in cache. => Nutanix pc client is not found in cache.

controllers/nutanixcluster_controller.go
```diff
@@ -178,12 +182,21 @@ func (r *NutanixClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			v3Client, err = CreateNutanixClient(ctx, r.SecretInformer, r.ConfigMapInformer, cluster)
+			if err != nil {
+				conditions.MarkFalse(cluster, infrav1.PrismCentralClientCondition, infrav1.PrismCentralClientInitializationFailed, capiv1.ConditionSeverityError, err.Error())
+				return ctrl.Result{Requeue: true}, fmt.Errorf("nutanix client error: %v", err)
```
deepakm-ntnx, line 192: we should add a time interval of how often should k8s control loop call back due to error.

controllers/nutanixmachine_controller.go
```diff
@@ -232,12 +233,22 @@ func (r *NutanixMachineReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			v3Client, err = CreateNutanixClient(ctx, r.SecretInformer, r.ConfigMapInformer, ntxCluster)
+			if err != nil {
+				conditions.MarkFalse(ntxMachine, infrav1.PrismCentralClientCondition, infrav1.PrismCentralClientInitializationFailed, capiv1.ConditionSeverityError, err.Error())
+				return ctrl.Result{Requeue: true}, fmt.Errorf("nutanix client error: %v", err)
+			}
+			prismclientcache.DefaultCache.Set(cluster.Name, v3Client)
+			conditions.MarkTrue(ntxMachine, infrav1.PrismCentralClientCondition)
+		} else {
+			return ctrl.Result{}, err
```
deepakm-ntnx, line 248: same comment as above

controllers/nutanixcluster_controller.go
```diff
@@ -178,12 +182,21 @@ func (r *NutanixClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 	conditions.MarkTrue(cluster, infrav1.CredentialRefSecretOwnerSetCondition)
 
-	v3Client, err := CreateNutanixClient(ctx, r.SecretInformer, r.ConfigMapInformer, cluster)
+	v3Client, err := prismclientcache.DefaultCache.Get(cluster.Name)
```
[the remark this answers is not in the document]
thunderboltsid: > @sid thanks for fixing this. one question, would it make sense to have per PC instead of per workload cluster client cache? sure we will have to implement ref count so that pc does not get removed if used by cluster. also understand that unique key for pc would be an issue if fqdn vs ip was provided by 2 separate clusters but that can may be fixed with get ip if fqdn is and store that? this will reduce the connection count to less as pc count will be less as compared to cluster count

No it doesn't make sense to have per PC as the users can be different per workload cluster even if the PC is the same. Overoptimization is the root of all evil; Let's stay away from that.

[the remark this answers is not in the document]
deepakm-ntnx: sure username ca be added as a part of connection too. only concern is from IAM side whether it can handle say 200 clusters having one connection per cluster. we can take it in separate PR if needed

controllers/nutanixcluster_controller.go
thunderboltsid: That's a good point, an exponential backoff might make more sense (albeit until a certain ceiling). Although, I'd stay away from making a change about it in this PR to keep the scope limited to introducing a cache.

deepakm-ntnx: this will be needed sooner than later to exponentially backoff given if user gives a wrong password then we will lock the pc due to too many calls with wrong creds

controllers/nutanixcluster_controller.go
thunderboltsid: Good catch. Added a delete from cache towards the end of reconcileDelete.

controllers/nutanixcluster_controller.go
thunderboltsid: fair enough. Updated the `Reconcile: true` calls to `ReconcileAfter: fiveSeconds`

deepakm-ntnx (approved):
lets create trackers for any missing things per comments

thunderboltsid:
> lets create trackers for any missing things per comments

@deepakm-ntnx this PR is atm not ready to merge. Part of the change here is pulling in an unreleased version of the prism-go-client. Once tests here pass, I'll create a PR for the prism-go-client. Once we get that PR merged and a new prism-go-client release out, then we can update the dependency in this PR to a proper version and get it merged.

deepakm-ntnx:
fyi: we may need to explore any of these internal connection pooling options https://www.loginradius.com/blog/engineering/tune-the-go-http-client-for-high-performance/#:~:text=By%20default%2C%20the%20Golang%20Http,(default%20is%2090%20seconds).
---
By default, the Golang Http client performs the connection pooling. When the request completes, that connection remains open until the idle connection timeout (default is 90 seconds). If another request came, that uses the same established connection instead of creating a new connection, after the idle connection time, the connection will return to the pool.

Using the connection pooling will keep less connection open and more requests will be served with minimal server resources.

When we not defined transport in the http.Client, it uses the default transport [Go HTTP Transport](https://golang.org/src/net/http/transport.go)

PR #402, opened 2024-04-02
fix: set the kube-vip var to default value instead of env for clusterclass

deepakm-ntnx:
for timebeing am going to reset these variables to false as default so that kube-vip with CAREN can be up . 

Future
once we add above things, we can make them configurable

- either make entire yaml as variable or
- allow configuring specific variables


dkoshkin (approved):

thunderboltsid (approved):
/lgtm
/approve

dkoshkin (approved):

PR #403, opened 2024-04-03
Ensure fallback config is read only when prismCentral is absent

deepakm-ntnx:
Skip reading fallback config file from `/etc/nutanix/config/prismCentral`
if NutanixCluster has prismCentral set.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -108,15 +108,15 @@ func (n *NutanixClientHelper) buildManagementEndpoint(ctx context.Context, nutan
 		providers = append(providers, providerForNutanixCluster)
 	} else {
 		log.Info(fmt.Sprintf("[WARNING] prismCentral attribute was not set on NutanixCluster %s in namespace %s. Defaulting to CAPX manager credentials", nutanixCluster.Name, nutanixCluster.Namespace))
-	}
-
-	// Fallback to building a provider using the global CAPX manager credentials
-	providerForLocalFile, err := n.buildProviderFromFile()
-	if err != nil {
-		return nil, fmt.Errorf("error building an environment provider from file: %w", err)
-	}
-	if providerForLocalFile != nil {
-		providers = append(providers, providerForLocalFile)
+		// Fallback to building a provider using prism central information from the CAPX management cluster
+		// using information from /etc/nutanix/config/prismCentral
+		providerForLocalFile, err := n.buildProviderFromFile()
+		if err != nil {
+			return nil, fmt.Errorf("error building an environment provider from file: %w", err)
+		}
+		if providerForLocalFile != nil {
+			providers = append(providers, providerForLocalFile)
+		}
 	}
 
 	// Initialize environment with providers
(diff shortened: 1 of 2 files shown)


pkg/client/client.go
```diff
@@ -102,21 +102,22 @@ func (n *NutanixClientHelper) buildManagementEndpoint(ctx context.Context, nutan
...
+		err = fmt.Errorf("[WARNING] prismCentral attribute was not set on NutanixCluster %s in namespace %s. Defaulting to CAPX manager credentials", nutanixCluster.Name, nutanixCluster.Namespace)
+	} else if providerForNutanixCluster != nil {
 		providers = append(providers, providerForNutanixCluster)
-	} else {
-		log.Info(fmt.Sprintf("[WARNING] prismCentral attribute was not set on NutanixCluster %s in namespace %s. Defaulting to CAPX manager credentials", nutanixCluster.Name, nutanixCluster.Namespace))
 	}
 
-	// Fallback to building a provider using the global CAPX manager credentials
-	providerForLocalFile, err := n.buildProviderFromFile()
 	if err != nil {
```
dkoshkin, line 112: I don't think we want to fallback if the user tried to provide creds and it was invalid (ie. returned an err)
IMO the fallback should only happen if the creds are nil, maybe something like https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/compare/main...dkoshkin:cluster-api-provider-nutanix:dkoshkin/fix-client-always-reading-from-file?expand=1

deepakm-ntnx: we dont want to return from first err != nil unless we try local file reading so in your fix. line 105.

thunderboltsid: I concur with @dkoshkin  here. The fallback should only happen in a nil, nil return case. If there's an error it needs to be propagated and not silenced.

pkg/client/client.go
```diff
@@ -106,17 +106,23 @@ func (n *NutanixClientHelper) buildManagementEndpoint(ctx context.Context, nutan
 	}
 	if providerForNutanixCluster != nil {
 		providers = append(providers, providerForNutanixCluster)
+		mptr, e1 := providerForNutanixCluster.GetManagementEndpoint(envTypes.Topology{})
```
deepakm-ntnx, line 109: why do we need to do it here, line 131 is already doing it rite?

deepakm-ntnx: pls see link mentioned by @dkoshkin in the conversation.

dkoshkin: Please see this and include the unit test for the changes https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/compare/main...dkoshkin:cluster-api-provider-nutanix:dkoshkin/fix-client-always-reading-from-file?expand=1

deepakm-ntnx: also pls add more unit tests if not already getting tested

thunderboltsid, line 109: Any reason why we are using `e1` instead of just `err`?

dkoshkin (approved):

adiantum (approved):
/lgtm
/approve

PR #404, opened 2024-04-04
Update additionalCategories clusterclass patch to target controlplane and workers separately

thunderboltsid:
This allows setting additionalCategories independently for controlplane and worker machine deployments.


deepakm-ntnx (approved):
we might need to update tests?

thunderboltsid:
> we might need to update tests?

no. that's the beauty of flexible tests.

adiantum (approved):
/lgtm
/approve

PR #407, opened 2024-04-08
Update release artifact workflow

thunderboltsid:
release workflows are missing the devbox vendored ko binary.

The release workflow was using `ko` but didn't have `debox run ---` before the `ko build` command leading to the workflow failing with `ko: command not found`


adiantum (approved):
/lgtm
/approve

PR #412, opened 2024-04-16
Update go directive in go.mod to 1.21

thunderboltsid:
Based on https://endoflife.date/go go 1.21 is the lowest still supported version of go. Go EOL policy is basically supporting two recent minor versions. Currently, CAPX main branch and 1.4.x releases have go directive set to 1.22 which is fine as long as we build it but is a problem when someone tries to build our repo with a supported but not latest version of go (e.g. Openshift building CAPX with their go 1.21 toolchain). We should change our go version to 1.21 and keep it set to the lowest supported version of go. i.e. we should stick to 1.21 until 1.23 is released and then move to 1.22 once 1.23 gets released and so on.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,11 @@
 module github.com/nutanix-cloud-native/cluster-api-provider-nutanix
 
-go 1.22.1
+// Based on https://endoflife.date/go go 1.21 is the lowest still supported version of go.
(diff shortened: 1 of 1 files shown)


thunderboltsid:
The template tests seem to break when changing the go version down to go1.21. This needs to be looked into in a bit more detail.

PR #400, opened 2024-03-22
removed optional annotation from non-optional fields

deepakm-ntnx:
**What this PR does / why we need it**:
removed optional annotation from non-optional fields 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**How Has This Been Tested?**:
CI tests

_Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and test output_

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

PR #414, opened 2024-04-21
Revert "removed optional annotation from non-optional fields (#400)"

thunderboltsid:
This reverts commit f6e079c5dd386256a27974338a111da3b628fb9b.

With https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/400 we made prismCentral a mandatory field in NutanixCluster. As a result, the test is now failing on both periodic jobs since and presubmits on new PRs with the following errors.

```
Expected success, but got an error:
    <*errors.StatusError | 0xc000411cc0>: 
    NutanixCluster.infrastructure.cluster.x-k8s.io "cluster-ntnx-client-qwqlsf" is invalid: spec.prismCentral: Required value
    {
        ErrStatus: {
            TypeMeta: {Kind: "", APIVersion: ""},
            ListMeta: {
                SelfLink: "",
                ResourceVersion: "",
                Continue: "",
                RemainingItemCount: nil,
            },
            Status: "Failure",
            Message: "NutanixCluster.infrastructure.cluster.x-k8s.io \"cluster-ntnx-client-qwqlsf\" is invalid: spec.prismCentral: Required value",
            Reason: "Invalid",
            Details: {
                Name: "cluster-ntnx-client-qwqlsf",
                Group: "infrastructure.cluster.x-k8s.io",
                Kind: "NutanixCluster",
                UID: "",
                Causes: [
                    {
                        Type: "FieldValueRequired",
                        Message: "Required value",
                        Field: "spec.prismCentral",
                    },
                ],
                RetryAfterSeconds: 0,
            },
            Code: 422,
        },
    } failed [FAILED] Timed out after 300.000s.
```

The clusterclass tests are also failing because variables are used to specify values for prismCentral and controlPlaneEndpoint in a a cluster created with a clusterclass reference and clusterclass patches patch those values into the cluster which was not accounted for when making the change. 

```
Running kubectl apply --kubeconfig /var/folders/g3/lb827bt96z10xz_m_c2xn93w0000gp/T/e2e-kind1856394220 -f -
stderr:
The NutanixClusterTemplate "nutanix-quick-start-nct" is invalid:
* spec.template.spec.controlPlaneEndpoint: Required value
* spec.template.spec.prismCentral: Required value
```

This PR reverts the changes made in https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/400

### DIFF
--- a/api/v1beta1/nutanixcluster_types.go
+++ b/api/v1beta1/nutanixcluster_types.go
@@ -46,12 +46,14 @@ type NutanixClusterSpec struct {
 
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
 	// host can be either DNS name or ip address
+	// +optional
 	ControlPlaneEndpoint capiv1.APIEndpoint `json:"controlPlaneEndpoint"`
 
 	// prismCentral holds the endpoint address and port to access the Nutanix Prism Central.
 	// When a cluster-wide proxy is installed, by default, this endpoint will be accessed via the proxy.
 	// Should you wish for communication with this endpoint not to be proxied, please add the endpoint to the
 	// proxy spec.noProxy list.
+	// +optional
 	PrismCentral *credentialTypes.NutanixPrismEndpoint `json:"prismCentral"`
 
 	// failureDomains configures failure domains information for the Nutanix platform.
@@ -60,7 +62,7 @@ type NutanixClusterSpec struct {
 	// +listType=map
 	// +listMapKey=name
 	// +optional
-	FailureDomains []NutanixFailureDomain `json:"failureDomains,omitempty"`
+	FailureDomains []NutanixFailureDomain `json:"failureDomains"`
 }
 
 // NutanixClusterStatus defines the observed state of NutanixCluster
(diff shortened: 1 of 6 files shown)


jimmidyson:
if the fields are indeed required, then they should remain as required in the underlying type imo.

For other providers that have required fields in cluster class resources that we add patches for to fill in required information (e.g. machine type in CAPA clusters), we use a dummy value `PLACEHOLDER` in the cluster class resources that makes the resource valid via the API. We then ensure that the variable used in the CC patch is required to ensure that the patch is applied and this dummy value is replaced.

I think this is worth an upstream discussion as it's a common use case, but I don't think reverting the previous PR here is actually the right thing to do.

tuxtof (approved):
LGTM

yannickstruyf3 (approved):

deepakm-ntnx:
@thunderboltsid can we post the exact test which fails since i would preferred to fix that test case by passing the required values instead of reverting the change. change has actually done what it is implemented for.

thunderboltsid:
> if the fields are indeed required, then they should remain as required in the underlying type imo.
> 
> For other providers that have required fields in cluster class resources that we add patches for to fill in required information (e.g. machine type in CAPA clusters), we use a dummy value `PLACEHOLDER` in the cluster class resources that makes the resource valid via the API. We then ensure that the variable used in the CC patch is required to ensure that the patch is applied and this dummy value is replaced.
> 
> I think this is worth an upstream discussion as it's a common use case, but I don't think reverting the previous PR here is actually the right thing to do.

Reverting the PR is done to fix the broken CI so we can unblock other changes until we fix this in a better tested manner.

thunderboltsid:
> @thunderboltsid can we post the exact test which fails since i would preferred to fix that test case by passing the required values instead of reverting the change. change has actually done what it is implemented for.

All clusterclass tests.

PR #415, opened 2024-04-23
Introduce a nutanix prism client cache

thunderboltsid:
During a recent incident, it was observed that creating a new Nutanix client for each request implies basic authentication for every request. This places unnecessary stress on IAM services. This stress was particularly problematic when the IAM services were already in a degraded state, thereby prolonging recovery efforts. Each basic auth request gets processed through the entire IAM stack, significantly increasing the load and impacting performance.

It's recommend that the client use session-auth cookies instead of basic auth for requests to Prism Central where possible. Given how the CAPX controller works currently, a new client is created per reconcile cycle. In https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/398 we switched to using Session-Auth instead of Basic-Auth. However, switching from Basic-Auth to Session-Auth alone wouldn’t solve the problem of consistent Basic-Auth calls. This is because each time a client is created, which is every reconcile cycle, it will still result in one Basic-Auth call to `/users/me` to fetch the session cookie. To alleviate this, we are adding a cache of clients and reusing the client from the cache across reconciliation cycles of the same cluster for both the NutanixCluster and NutanixMachine reconciliation.

Screenshot 2024-05-03 at 15 16 31

In a large-scale setup of 40+ clusters w/ 4 nodes each, we were able to see a noticeable drop in QPS to the IAM stack for the `oidc/token` calls. Before the client caching, a controller restart led to 10+ QPS on `oidc/token` endpoint with a steady state at around 0.5 QPS. After deploying the client cache changes, we saw a peak of ~3 QPS as caches warmed up and dropped to 0 QPS afterwards with sporadic requests only when session token refresh was needed.  As we can see, with the changes proposed in this document, we were able to reduce the number of high-impact calls to IAM significantly.


controllers/nutanixcluster_controller.go
```diff
@@ -170,20 +170,11 @@ func (r *NutanixClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		log.V(1).Info(fmt.Sprintf("Patched NutanixCluster. Status: %+v", cluster.Status))
 	}()
 
-	err = r.reconcileCredentialRef(ctx, cluster)
```
dkoshkin, line 173: What do you think about keeping `r.reconcileCredentialRef` but also returning the `Secret` and then passing that Secret into `getPrismCentralClientForCluster`, that would make the function a little more focused and more unit testable?

It would also make the changes in this PR smaller.

(originally left this comment in https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/398#discussion_r1575308135)

thunderboltsid: as discussed offline, the `reconcileCredentialRef` is getting called from both machine and cluster controller and as such can't be a method on the cluster reconciler.

controllers/helpers.go
```diff
@@ -760,3 +759,92 @@ func GetFailureDomain(failureDomainName string, nutanixCluster *infrav1.NutanixC
...
+
+	if err := reconcileCredentialRef(ctx, k8sClient, cluster); err != nil {
+		log.Error(err, fmt.Sprintf("error occurred while reconciling credential ref for cluster %s", capiCluster.Name))
+		conditions.MarkFalse(cluster, infrav1.CredentialRefSecretOwnerSetCondition, infrav1.CredentialRefSecretOwnerSetFailed, capiv1.ConditionSeverityError, err.Error())
+		return nil, err
+	}
+	conditions.MarkTrue(cluster, infrav1.CredentialRefSecretOwnerSetCondition)
+
+	clientHelper := nutanixClient.NewHelper(secretInformer, mapInformer)
+	mep, err := clientHelper.BuildManagementEndpoint(ctx, cluster)
```
dkoshkin, line 831: nit, is `mep` already used in the codebase? Maybe just `managementEndpoint`

pkg/client/cache.go
```diff
@@ -0,0 +1,26 @@
...
+
+// Key is the namespace/name of te NutanixCluster CR
+func (c *CacheParams) Key() string {
+	return c.NutanixCluster.GetNamespacedName()
+}
+
+// ManagementEndpoint returns the management endpoint of the NutanixCluster CR
+func (c *CacheParams) ManagementEndpoint() types.ManagementEndpoint {
+	return *c.PrismManagementEndpoint
+}
```
dkoshkin, line 26: Great interface in the client!

controllers/helpers_test.go
```diff
@@ -18,16 +18,27 @@ package controllers
...
 	"sigs.k8s.io/cluster-api/util"
+	ctlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	infrav1 "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/api/v1beta1"
-
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
+	mockctlclient "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/mocks/ctlclient"
+	mockk8sclient "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/mocks/k8sclient"
+	nutanixClient "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkg/client"
```
dkoshkin, line 41: nit
```suggestion
	nutanixclient "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkg/client"
```

controllers/nutanixcluster_controller.go
```diff
@@ -44,6 +43,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	infrav1 "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/api/v1beta1"
+	nutanixClient "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkg/client"
```
dkoshkin, line 46: >They are lower case, with no under_scores or mixedCaps.

nit: https://go.dev/blog/package-names
```suggestion
	nutanixclient "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkg/client"
```

dkoshkin (approved):
Thanks for all the new tests!

pkg/client/cache.go
thunderboltsid: Kudos to @adiantum for the initial idea

adiantum: Thank you Mr. @thunderboltsid

adiantum (approved):
Overall looks good, but I think we should plan some refactoring in separate PR. Please consider my comments.
/lgtm
/approve

controllers/nutanixcluster_controller.go
thunderboltsid: With the responsibility of providing the secret contents management at any time shifted to the client cache itself, this no longer needs to be called by machine reconciler. I've moved it back to it's original role as a method on cluster reconciler.

adiantum (approved):
/lgtm
/approve

PR #416, opened 2024-04-23
Explicitly set a rate limiter for controller config

thunderboltsid:
The rate limiter is a MaxOfRateLimiter similar to the default one. However, now the base delay, max delay, bucket size, and qps can be configured using flags. The default values are 500ms for base delay and 15min for max delay in the exponential backoff limiter (custom values) and 10 qps and 100 bucket size for the bucket limiter (same as the default one).

### DIFF
--- a/controllers/options.go
+++ b/controllers/options.go
@@ -1,10 +1,15 @@
 package controllers
 
-import "errors"
+import (
+	"errors"
+
+	"k8s.io/client-go/util/workqueue"
+)
 
 // ControllerConfig is the configuration for cluster and machine controllers
 type ControllerConfig struct {
 	MaxConcurrentReconciles int
+	RateLimiter             workqueue.RateLimiter
 }
 
 // ControllerConfigOpts is a function that can be used to configure the controller config
@@ -17,6 +22,17 @@ func WithMaxConcurrentReconciles(max int) ControllerConfigOpts {
 			return errors.New("max concurrent reconciles must be greater than 0")
 		}
 		c.MaxConcurrentReconciles = max
+		return nil
+	}
+}
+
+// WithRateLimiter sets the rate limiter for the controller
+func WithRateLimiter(rateLimiter workqueue.RateLimiter) ControllerConfigOpts {
+	return func(c *ControllerConfig) error {
+		if rateLimiter == nil {
+			return errors.New("rate limiter cannot be nil")
+		}
+		c.RateLimiter = rateLimiter
 		return nil
 	}
 }
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -26,6 +27,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/spf13/pflag"
 	"go.uber.org/zap/zapcore"
+	"golang.org/x/time/rate"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/informers"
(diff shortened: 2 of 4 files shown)


deepakm-ntnx:
just documenting here for completeness. https://stuartleeks.com/posts/error-back-off-with-controller-runtime/

thunderboltsid:
> just documenting here for completeness. https://stuartleeks.com/posts/error-back-off-with-controller-runtime/

This article is slightly outdated. Newer versions of the controller-runtime use workqueue.DefaultControllerRateLimiter provided by client-go library by default instead.

dkoshkin (dismissed):

main.go
```diff
@@ -88,6 +97,13 @@ func parseFlags(config *managerConfig) {
 	pflag.IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", defaultMaxConcurrentReconciles,
 		"The maximum number of allowed, concurrent reconciles.")
 
+	var baseDelay, maxDelay time.Duration
+	var bucketSize, qps int
+	pflag.DurationVar(&baseDelay, "rate-limiter-base-delay", 500*time.Millisecond, "The base delay for the rate limiter.")
+	pflag.DurationVar(&maxDelay, "rate-limiter-max-delay", 15*time.Minute, "The maximum delay for the rate limiter.")
```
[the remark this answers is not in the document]
thunderboltsid: 15m is the session token expiration time. If you do the math, the longest time it will actually use that is less than 15m would be 512seconds (8ish mins)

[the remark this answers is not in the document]
thunderboltsid: will point you to an internal google doc.

dkoshkin (approved):

controllers/options.go
```diff
@@ -20,3 +25,14 @@ func WithMaxConcurrentReconciles(max int) ControllerConfigOpts {
 		return nil
 	}
 }
+
+// WithRateLimiter sets the rate limiter for the controller
+func WithRateLimiter(rateLimiter workqueue.RateLimiter) ControllerConfigOpts {
+	return func(c *ControllerConfig) error {
+		if rateLimiter == nil {
+			return errors.New("rate limiter cannot be nil")
```
deepakm-ntnx, line 33: nit: full stop at the end of the statement.

main.go
```diff
@@ -88,6 +98,13 @@ func parseFlags(config *managerConfig) {
...
+	pflag.IntVar(&bucketSize, "rate-limiter-bucket-size", 100, "The bucket size for the rate limiter.")
+	pflag.IntVar(&qps, "rate-limiter-qps", 10, "The QPS for the rate limiter.")
```
deepakm-ntnx, line 106: what happens when someone passes it as zero? should we be setting a min qps as 1 atleast else not sure what happens

main.go
```diff
@@ -189,6 +207,7 @@ func runManager(ctx context.Context, mgr manager.Manager, config *managerConfig)
 
 	clusterControllerOpts := []controllers.ControllerConfigOpts{
 		controllers.WithMaxConcurrentReconciles(config.concurrentReconcilesNutanixCluster),
+		controllers.WithRateLimiter(config.rateLimiter),
```
deepakm-ntnx, line 210: https://release-1-5.cluster-api.sigs.k8s.io/developer/architecture/controllers/tuning. will this also impact api server calls?

deepakm-ntnx (approved):

deepakm-ntnx:
pls create a tracker to test the behavior in general either manual or automated

controllers/options.go
thunderboltsid: This is usually inadvisable in Go.

main.go
thunderboltsid: I'll add some variable validation and unit tests for them.

PR #420, opened 2024-04-25
Enable jobs to run on release-v1.2

thunderboltsid:
Originally, there was no release branch for v1.2.x releases. Each release was cut from the main. As a result jobs were targeting main branch only. Now that we have created a release-v1.2 branch for backporting some patches we need to manually enable jobs in the branch to target the release branch.


yannickstruyf3 (approved):

tuxtof (approved):
LGTM

PR #418, opened 2024-04-24
Cherry-pick PR #354, #355, #360, and #398 to release-v1.2

thunderboltsid:
### Cherry-Pick Details
- **Original PR Titles:**
    * fix: improve error handling for Prism Client
    * refactor: task status file
    * refactor: client.go file helper methods
    * Switch Nutanix Client to use Session Auth instead of Basic Auth
- **Original PR URLs:**
    * https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/354
    * https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/355
    * https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/360
    * https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/398
- **Original Commit SHAs:**
    * 1f8716d88ee3ef965599254144cd2ce2a4c5e37c
    * 5fd7ada9b99ab7f3b2b48dd15ae186c017296c4b
    * 3fa9d45b680065e2471e22854189e1ff1d6b2eb0
    * 03c61d34afe159fe91940724e44f821dfeed1d1c
- **Cherry-Pick Reason:** Cherry-picking changes that allow us to cleanly cherry-pick new patches related to session auth

PR #425, opened 2024-04-30
Introduce a nutanix prism client cache 

thunderboltsid:
During a recent incident, it was observed that creating a new Nutanix client for each request implies basic authentication for every request. This places unnecessary stress on IAM services. This stress was particularly problematic when the IAM services were already in a degraded state, thereby prolonging recovery efforts. Each basic auth request gets processed through the entire IAM stack, significantly increasing the load and impacting performance.

It's recommend that the client use session-auth cookies instead of basic auth for requests to Prism Central where possible. Given how the CAPX controller works currently, a new client is created per reconcile cycle. In https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/418 we switched to using Session-Auth instead of Basic-Auth. However, switching from Basic-Auth to Session-Auth alone wouldn’t solve the problem of consistent Basic-Auth calls. This is because each time a client is created, which is every reconcile cycle, it will still result in one Basic-Auth call to `/users/me` to fetch the session cookie. To alleviate this, we are adding a cache of clients and reusing the client from the cache across reconciliation cycles of the same cluster for both the NutanixCluster and NutanixMachine reconciliation.

Screenshot 2024-05-03 at 15 16 31

In a large-scale setup of 40+ clusters w/ 4 nodes each, we were able to see a noticeable drop in QPS to the IAM stack for the `oidc/token` calls. Before the client caching, a controller restart led to 10+ QPS on `oidc/token` endpoint with a steady state at around 0.5 QPS. After deploying the client cache changes, we saw a peak of ~3 QPS as caches warmed up and dropped to 0 QPS afterwards with sporadic requests only when session token refresh was needed.  As we can see, with the changes proposed in this document, we were able to reduce the number of high-impact calls to IAM significantly.


deepakm-ntnx (approved):

thunderboltsid:
Closed in favor of https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/430

PR #436, opened 2024-05-27
Add a watch on CAPI cluster in NutanixMachineReconciler

thunderboltsid:
- Refactor SetupWithManager to call `Build` explicitly to add a watch to the controller after `Build` is called instead of calling `Complete`.
- Add a watch to the CAPI cluster with a `ClusterUnpausedAndInfrastructureReady` predicate
- Ensure `RateLimiter` gets correctly propagated to the underlying `controller.Options`. Earlier `controllerConfig.Ratelimiter` was not passed in the `controller.Options` struct passed to the controller builder


thunderboltsid:
Change seems to be failing on lint issues that have been fixed in https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/437

adiantum (approved):
/lgtm
/approve

deepakm-ntnx:
> * Ensure `RateLimiter` gets correctly propagated to the underlying `controller.Options`

Can we document how did we come to conclusion that this was not propagating correctly?

thunderboltsid:
> > * Ensure `RateLimiter` gets correctly propagated to the underlying `controller.Options`
> 
> Can we document how did we come to conclusion that this was not propagating correctly?

By looking at the code and realizing that the controllerconfig.Ratelimiter was not passed in the controller.Options struct passed to the controller builder.

controllers/nutanixmachine_controller_test.go
```diff
@@ -228,3 +236,133 @@ func TestNutanixMachineReconciler(t *testing.T) {
...
+		controllerConfig: &ControllerConfig{
+			MaxConcurrentReconciles: 1,
+		},
+	}
+
+	err = reconciler.SetupWithManager(ctx, mgr)
+	assert.Error(t, err)
+}
+
+func TestNutanixMachineReconciler_SetupWithManager_ClusterToTypedObjectsMapperError(t *testing.T) {
```
deepakm-ntnx, line 326: comments on the purpose of the test would be appreciated.

deepakm-ntnx (approved):

controllers/nutanixmachine_controller_test.go
thunderboltsid: it's a test for the case when ClusterToTypedObjectsMapper call fails inside SetupWithManager. Largely added to appease the codecov check.

PR #438, opened 2024-05-28
Set OwnerRef on TrustBundle configmaps and add selfhosted E2E test

thunderboltsid:
This PR adds a E2E test for clusterctl move and also adds a method to reconcile trust bundle refs in NutanixmachineReconciler to  set ownerReference on trust bundle configmaps so those configmaps get moved when clusterctl move operation happens.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,18 +4,17 @@ GOTEST=$(GOCMD) test
 GOGET=$(GOCMD) get
 GOTOOL=$(GOCMD) tool
 EXPORT_RESULT?=false # for CI please set EXPORT_RESULT to true
-# Image URL to use all building/pushing image targets
-IMG ?= ghcr.io/nutanix-cloud-native/cluster-api-provider-nutanix/controller:latest
+
+GIT_COMMIT_HASH=$(shell git rev-parse HEAD)
+LOCAL_IMAGE_REGISTRY ?= localhost:5000
+IMG_REPO=${LOCAL_IMAGE_REGISTRY}/cluster-api-provider-nutanix
+IMG_TAG=e2e-${GIT_COMMIT_HASH}
+MANAGER_IMAGE=${IMG_REPO}:${IMG_TAG}
 
 # Extract base and tag from IMG
-IMG_REPO ?= $(word 1,$(subst :, ,${IMG}))
-IMG_TAG ?= $(word 2,$(subst :, ,${IMG}))
 LOCAL_PROVIDER_VERSION ?= ${IMG_TAG}
-ifeq (${IMG_TAG},)
-IMG_TAG := latest
-endif
 
-ifeq (${LOCAL_PROVIDER_VERSION},latest)
+ifeq (${LOCAL_PROVIDER_VERSION},${IMG_TAG})
 # TODO(release-blocker): Change this versions after release when required here and in e2e config (test/e2e/config/nutanix.yaml)
 LOCAL_PROVIDER_VERSION := v1.4.99
 endif
@@ -62,21 +61,6 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-# Get latest git hash
-GIT_COMMIT_HASH=$(shell git rev-parse HEAD)
-
-# Get the local image registry required for clusterctl upgrade tests
-LOCAL_IMAGE_REGISTRY ?= localhost:5000
-
-ifeq (${MAKECMDGOALS},test-e2e-clusterctl-upgrade)
-	IMG_TAG=e2e-${GIT_COMMIT_HASH}
-	IMG_REPO=${LOCAL_IMAGE_REGISTRY}/controller
-endif
-
-ifeq (${MAKECMDGOALS},docker-build-e2e)
-	IMG_TAG=e2e-${GIT_COMMIT_HASH}
-endif
-
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # This is a requirement for 'setup-envtest.sh' in the test target.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -140,6 +125,11 @@ help: ## Display this help.
 manifests: ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	controller-gen $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
+.PHONY: set-manifest-image
+set-manifest-image: ## Update kustomize image patch file for default resource.
+	$(info Updating kustomize image patch file for default resource)
+	sed -i'' -e 's@image: .*@image: '"${MANAGER_IMAGE}"'@' ./config/manager/manager_image_patch.yaml
+
```
thunderboltsid, line 132: target no longer used, can be removed

adiantum (approved):
/lgtm
/approve

config/manager/manager_image_patch.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+        # Change the value of image field below to your controller image URL
+        - image: ghcr.io/nutanix-cloud-native/cluster-api-provider-nutanix/controller:latest
```
deepakm-ntnx, line 11: can we make sure to remove the latest tagged image in github registry? else sometimes local dev pull that if we have not done push kind https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pkgs/container/cluster-api-provider-nutanix%2Fcontroller/216366670?tag=latest

controllers/nutanixcluster_controller.go
```diff
@@ -178,6 +179,13 @@ func (r *NutanixClusterReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 	conditions.MarkTrue(cluster, infrav1.CredentialRefSecretOwnerSetCondition)
 
+	if err := r.reconcileTrustBundleRef(ctx, cluster); err != nil {
+		log.Error(err, fmt.Sprintf("error occurred while reconciling trust bundle ref for cluster %s", capiCluster.Name))
+		conditions.MarkFalse(cluster, infrav1.TrustBundleSecretOwnerSetCondition, infrav1.TrustBundleSecretOwnerSetFailed, capiv1.ConditionSeverityError, err.Error())
+		return reconcile.Result{}, err
+	}
+	conditions.MarkTrue(cluster, infrav1.TrustBundleSecretOwnerSetCondition)
```
deepakm-ntnx, line 187: we should not set this if we never used trust bundle.  err nil is returned in following case as well <pre>
	trustBundleRef := nutanixCluster.GetPrismCentralTrustBundle()
	if trustBundleRef == nil {
		log.Info(fmt.Sprintf("trust bundle ref is nil for cluster %s", nutanixCluster.Name))
		return nil
</pre>

Makefile
```diff
@@ -62,21 +61,6 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-# Get latest git hash
-GIT_COMMIT_HASH=$(shell git rev-parse HEAD)
-
-# Get the local image registry required for clusterctl upgrade tests
-LOCAL_IMAGE_REGISTRY ?= localhost:5000
-
-ifeq (${MAKECMDGOALS},test-e2e-clusterctl-upgrade)
```
deepakm-ntnx, line 71: have we tested test-e2e-clusterctl-upgrade to make sure we are not breaking it with this change

test/e2e/clusterctl_move_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	capie2e "sigs.k8s.io/cluster-api/test/e2e"
+)
+
+var _ = Describe("[clusterctl-move] Create a Self-Hosted CAPX Cluster", Label("capx-feature-test", "clusterctl-move"), func() {
+	// Run the CAPI SelfHostedSpec test suite
+	// This test suite will create a self-hosted CAPX cluster
+	capie2e.SelfHostedSpec(ctx, func() capie2e.SelfHostedSpecInput {
```
deepakm-ntnx, line 29: we should add one for clusterclass based cluster move as well

Makefile
thunderboltsid: Yes, you can see the tests that run and have passed in the status section at the end the PR.

config/manager/manager_image_patch.yaml
thunderboltsid: no, we will not be removing the latest tagged image; that breaks away from convention.

controllers/nutanixcluster_controller.go
thunderboltsid: good point, changed the logic, although I moved the condition marking logic to the reconcileTrustBundleRef method to keep the logic cleaner.

config/manager/manager_image_patch.yaml
deepakm-ntnx: i have getting error on fresh make deploy on a kind cluster where capx controller crashes with invalid flag --metrics-... and expects --diagnostics..... i feel that its because of stale latest image in cloud. do you know if its getting updated at all?

deepakm-ntnx (approved):
we need to fix the codecov errors

config/manager/manager_image_patch.yaml
thunderboltsid: latest image in production is likely because 'latest' refers to 1.3.4. In 1.4.x (and main) we have different options i.e. diagnostics-addr. I'm wondering if it has something to do with that. In that case imo, it's make deploy target that needs fixing and not the image.

adiantum (approved):
/lgtm
/approve

PR #444, opened 2024-06-03
Ensure targets referring to IMG refer to MANAGER_IMAGE instead

thunderboltsid:
`IMG` target does not exist and those commands were fail.


adiantum (approved):
/lgtm
/approve

PR #446, opened 2024-06-03
Remove `listMapKey` for subnets list in failure domains

thunderboltsid:
A subnet can either be referenced as a UUID or as Name depending on the type of the reference. As a result we can't rely on either of them as a defacto map key.


dkoshkin (approved):
Yep, I had a similar comment https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/pull/684#discussion_r1612294262, didn't realize we had it in CAPX already.

Thanks for removing.

PR #455, opened 2024-07-09
Add the checkout ref for incorporating PR

thunderboltsid:
Since we moved to using pull_request_target which executes the workflow in the context of the base repository but the default HEAD that gets checked out is now the main branch rather than the PR. This ensures the pull request HEAD is checked out before running the workflow.


tuxtof (approved):
LGTM

adiantum (approved):

PR #463, opened 2024-07-27
Swap out busybox image to non-dockerhub

thunderboltsid:
This is to ensure fewer rate limit issues via dockerhub.


tuxtof:
@thunderboltsid best is use a more official busybox image and not a random one coming from Prom project

we can go with `cgr.dev/chainguard/busybox`

adiantum (approved):
lgtm

PR #465, opened 2024-07-30
fix(controller): Move v4 client instantiation to reconcileDelete

thunderboltsid:
In order to reduce the number of calls made to /prism_central to establish whether a v4 client can be created, we move the v4 client creation to reconcileDelete function as currently that's the only place where we need to make v4 VG detach calls. This ensures we don't impact the regular reconcile cycle and all the new calls only take place on the delete path.


controllers/nutanixmachine_controller.go
```diff
@@ -327,81 +311,109 @@ func (r *NutanixMachineReconciler) reconcileDelete(rctx *nctx.MachineContext) (r
...
+	vm, err := FindVMByUUID(ctx, v3Client, vmUUID)
+	// Error while finding VM
+	if err != nil {
+		errorMsg := fmt.Errorf("error finding vm %s with uuid %s: %v", vmName, vmUUID, err)
+		log.Error(errorMsg, "error finding vm")
+		conditions.MarkFalse(rctx.NutanixMachine, infrav1.VMProvisionedCondition, infrav1.DeletionFailed, capiv1.ConditionSeverityWarning, errorMsg.Error())
+		return reconcile.Result{}, errorMsg
+	}
+
+	// Vm not found
```
dkoshkin, line 333: nit: lets be consistent with "VM" or "vm" in all the comments and info and error messages
```suggestion
	// VM not found
```

controllers/nutanixmachine_controller.go
```diff
@@ -327,81 +311,109 @@ func (r *NutanixMachineReconciler) reconcileDelete(rctx *nctx.MachineContext) (r
...
+				vgDetachNeeded = true
+				break
 			}
+		}
+	}
 
-			// Delete the VM since the VM was found (err was nil)
-			deleteTaskUUID, err := DeleteVM(ctx, v3Client, vmName, vmUUID)
+	if vgDetachNeeded {
+		createV4Client, err := isPrismCentralV4Compatible(ctx, v3Client)
```
dkoshkin, line 385: Can we move this to a separate function and have it a little flatter?

controllers/nutanixmachine_controller.go
```diff
@@ -327,81 +311,109 @@ func (r *NutanixMachineReconciler) reconcileDelete(rctx *nctx.MachineContext) (r
...
-		// Search for VM by UUID
-		vm, err := FindVMByUUID(ctx, v3Client, vmUUID)
-		// Error while finding VM
+		log.Info(fmt.Sprintf("Removing finalizers for VM %s during delete reconciliation", vmName))
+		ctrlutil.RemoveFinalizer(rctx.NutanixMachine, infrav1.NutanixMachineFinalizer)
+		return reconcile.Result{}, nil
+	}
+
+	// Search for VM by UUID
+	vm, err := FindVMByUUID(ctx, v3Client, vmUUID)
```
dkoshkin, line 324: Its hard to read the diff from existing code, did any existing code change thats not related to the VG detachment?

thunderboltsid: I'd recommend reading the diff with the setting on whitespace removal separated. This PR tries to keep cyclomatic complexity in check by removing some outer loops in place.

controllers/nutanixmachine_controller.go
```diff
@@ -327,81 +311,112 @@ func (r *NutanixMachineReconciler) reconcileDelete(rctx *nctx.MachineContext) (r
...
 
-	return reconcile.Result{}, nil
+func (r *NutanixMachineReconciler) detachVolumeGroups(rctx *nctx.MachineContext, vmUUID string) error {
+	createV4Client, err := isPrismCentralV4Compatible(rctx.Context, rctx.NutanixClient)
+	if err != nil {
+		err := fmt.Errorf("error occurred while checking compatibility for Prism Central v4 APIs: %w", err)
+		return err
+	}
+
+	if createV4Client {
```
dkoshkin, line 406: nit: can flatten the function and move the business logic out of the if
```suggestion
	if !canCreateV4Client {
	  // Detaching Volume Groups is only supported with a v4 client.
	  // But this PC version does not support v4 APIs, return early.
	  return nil
	 }
```

dkoshkin (approved):

adiantum (approved):
lgtm

PR #468, opened 2024-08-02
fix(controller)!: Set UEFI as the default BootType

thunderboltsid:
As we have switched to UEFI as the default Boot Type in Prism Central, we should do the same with CAPX to stay aligned with the direction of the product.


tuxtof:
if we do that we need to switch to v2 and document/warn it properly as this is a breaking change for existing customer

PR #477, opened 2024-08-25
fix(controller): Optimize logic for detach VG from VM for cases with large number of VGs

thunderboltsid:
We optimize the logic and add some guardrails
* Add a timeout on the context for polling
* Increase the polling interval to 5s
* Set immediate polling to false to reduce likelihood of a 404 on tasks
* Call detachVG one after the other and build a list of tasks
* wait for all tasks in the list to reach a terminal state

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -20,24 +20,22 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/go-logr/logr"
 	"reflect"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/google/uuid"
 	"github.com/nutanix-cloud-native/prism-go-client/utils"
 	prismclientv3 "github.com/nutanix-cloud-native/prism-go-client/v3"
 	prismclientv4 "github.com/nutanix-cloud-native/prism-go-client/v4"
-	prismcommonconfig "github.com/nutanix/ntnx-api-golang-clients/prism-go-client/v4/models/common/v1/config"
 	prismapi "github.com/nutanix/ntnx-api-golang-clients/prism-go-client/v4/models/prism/v4/config"
 	prismconfig "github.com/nutanix/ntnx-api-golang-clients/volumes-go-client/v4/models/prism/v4/config"
 	volumesconfig "github.com/nutanix/ntnx-api-golang-clients/volumes-go-client/v4/models/volumes/v4/config"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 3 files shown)


controllers/helpers.go
```diff
@@ -54,7 +54,8 @@ const (
 
 	gpuUnused = "UNUSED"
 
-	pollingInterval = time.Second * 2
+	taskPollingInterval   = 5 * time.Second
+	taskCompletionTimeout = 5 * time.Minute
```
jimmidyson, line 58: Is this the maximum time a control loop could run for? This seems a long time, especially if we only have 10 workers, meaning we can only detach volumes from 10 machines at a time. I'm concerned about the effect of this when there are lots of machines, either as large cluster or for many clusters. This would potentially affect clusters coming up and being upgraded as well as being deleted right?

jimmidyson: Is there any way we can filter tasks using https://developers.nutanix.com/api-reference?namespace=prism&version=v4.0.b1#tag/Tasks/operation/listTasks so we don't have to wait here at all?

controllers/nutanixmachine_controller.go
```diff
@@ -362,7 +362,7 @@ func (r *NutanixMachineReconciler) reconcileDelete(rctx *nctx.MachineContext) (r
 		}
 		log.V(1).Info(fmt.Sprintf("no running tasks anymore... Initiating delete for VM %s with UUID %s", vmName, vmUUID))
 	} else {
-		log.V(1).Info(fmt.Sprintf("no task UUID found on VM %s. Starting delete.", *vm.Spec.Name))
+		log.V(1).Info(fmt.Sprintf("no task UUID found on VM %s. Starting delete.", vmName))
```
jimmidyson, line 365: Do detach VG tasks get associated with the VM? If so then we don't need to wait for tasks at all if we've fired off the VG detachments, do we?

controllers/helpers.go
thunderboltsid: let's avoid batching for now. I'd like to familiarize myself w/ those APIs a bit more before committing CAPX to rely on them.

thunderboltsid: as discussed offline, let's omit waiting altogether and just requeue for the time being

controllers/nutanixmachine_controller.go
thunderboltsid: it's a VG api call and not a VM api call. So, I'd be somewhat doubtful. I agree we don't need to wait for tasks for now. So, let's remove them as agreed offline.

jimmidyson (approved):

faiq (approved):

PR #479, opened 2024-08-26
feat(ccm): Update the CCM version to v0.4.1

thunderboltsid:
Also add a update-ccm target to automate updating CCM Version in future.


adiantum (approved):
lgtm

PR #480, opened 2024-08-26
fix(template): add CONTROL_PLANE_ENDPOINT_IP to ignoredNodeIPs list in CCM

tuxtof:

By adding CONTROL_PLANE_ENDPOINT_IP to the ignoredNodeIPs in the CCM config, we ensure
CCM will not assign the CONTROL_PLANE_ENDPOINT_IP as the node IP for any node. This could
previously happen as both CONTROL_PLANE_ENDPOINT_IP and node IP on an external DHCP network
would be learned IPs. In case of a slow DHCP assignment, the CONTROL_PLANE_ENDPOINT_IP
could be assigned as the first learned IP on the VM leading to potential issues when the CCM would then
assign the first IP as the node IP even though it was a floating IP.
It also causing issue with IPAM mode and third-party tools who are using IP from nodes

manual deployment of modified ressources

**Release note**:

```release-note
- Set CCM to ignore Control Plane IP
```

### DIFF
--- a/templates/ccm/nutanix-ccm.yaml
+++ b/templates/ccm/nutanix-ccm.yaml
@@ -41,7 +41,10 @@ data:
       "enableCustomLabeling": ${CCM_CUSTOM_LABEL=false},
       "topologyDiscovery": {
         "type": "Prism"
-      }
+      },
+      "ignoredNodeIPs": [
+        "${CONTROL_PLANE_ENDPOINT_IP}"
+      ]
     }
 ---
 # Source: nutanix-cloud-provider/templates/rbac.yaml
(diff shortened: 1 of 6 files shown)


thunderboltsid:
@tuxtof we're bumping up CCM in https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/479 already. Can you approve that one, rebase this and keep the scope limited to adding control plane endpoint IP to the ignore IPs list?

adiantum (approved):
lgtm

thunderboltsid:
We also need to update the template in test/e2e/data/infrastructure-nutanix/ccm-update.yaml

thunderboltsid (approved):

thunderboltsid (approved):

PR #484, opened 2024-10-02
fix(bug): Encode name in filter by name FIQL Query

thunderboltsid:
Not encoding the name can cause errors when non-standard characters are present in the name.

**How has this been tested?**
Unter testing
* Create a new subnet with escapable characters in the name: `vlan173|ncn|dev|sandbox`
* Create a new cluster with a machine deployment pointing to that subnet
* Observe the cluster come up successfully

Current status: with these changes we are still seeing failed machines with:
```
  failureMessage: 'Failure detected from referenced resource infrastructure.cluster.x-k8s.io/v1beta1,
    Kind=NutanixMachine with name "quick-start-0j5mcd-zb6l8-k6fqs": failed to retrieve
    subnet by name vlan173|ncn|dev|sandbox'
```

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


adiantum (approved):
lgtm
but you need to add some tests

jimmidyson (approved):

thunderboltsid:
Closing this PR as URL encoding doesn't take care of all escaping. This /subnets/list is a POST endpoint and the filter is passed as a json object where the `|` has to be escaped as `\\|`.  We have decided to make a tradeoff in favor correctness in exchange for performance and do all filtering client-side. See #485

PR #489, opened 2024-12-02
refactor: Make available complete image info, instead of UUID only

dlipovetsky:
**What this PR does / why we need it**:
We plan to make decisions based on the complete image info.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**How Has This Been Tested?**:

_Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and test output_

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

PR #498, opened 2025-01-16
feat: adds image templating

faiq:

This PR allows us to do templating for nutanix machines 

```
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: NutanixMachineTemplate
metadata:
  name: faiq-mt-0
  namespace: default
spec:
  template:
    spec:
      bootType: legacy
      cluster:
        name: ncn-dev-sandbox
        type: name
      imageLookup:
        baseOS: rhel-8.10
        format: nkp-{{.BaseOS}}-.*?{{.K8sVersion}}-.*
      memorySize: 4Gi
````

instead of providing an image name or uuid. this is a quality of life feature that makes it easier to create clusters

Fixes https://jira.nutanix.com/browse/NCN-101000

Run `make  test-e2e-calico  LABEL_FILTERS='image-templating'` to test the new e2e test i added to verify the behavior. 

**Release note**:

```release-note
feat: adds image lookup based on templating
```


templates/cluster-template.yaml
```diff
@@ -579,9 +579,8 @@ spec:
       cluster:
         name: ${NUTANIX_PRISM_ELEMENT_CLUSTER_NAME}
         type: name
-      image:
-        name: ${NUTANIX_MACHINE_TEMPLATE_IMAGE_NAME}
-        type: name
+      imageLookupBaseOS: ${BASE_OS}
+      imageLookupFormat: ${LOOKUP_FORMAT}
```
faiq, line 583: this probably shouldn't be in the base template - it causes the e2e tests fail. id like some guidance on how to deal with this.

controllers/helpers.go
```diff
@@ -299,45 +302,134 @@ func GetSubnetUUID(ctx context.Context, client *prismclientv3.Client, peUUID str
...
+	}
+	sorted := sortImages(responseImages.Entities)
+	if len(sorted) == 0 {
+		return "", fmt.Errorf("failed to find image with filter %s", filter)
+	}
+	return *sorted[0].Metadata.UUID, nil
+}
+
+func sortImages(images []*prismclientv3.ImageIntentResponse) []*prismclientv3.ImageIntentResponse {
+	sort.Slice(images, func(i, j int) bool {
```
supershal, line 390: can we sort by "created date" and use the latest one?

config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixmachines.yaml
```diff
@@ -187,6 +187,25 @@ spec:
                 required:
                 - type
                 type: object
+              imageLookupBaseOS:
```
supershal, line 190: wdyt moving it under a container so that it can be extended in future?
```yaml
imageLookup:
  baseOS:
  format:
```

controllers/helpers.go
```diff
@@ -299,45 +302,134 @@ func GetSubnetUUID(ctx context.Context, client *prismclientv3.Client, peUUID str
...
+	ctx context.Context,
+	client *prismclientv3.Client,
+	imageName,
+	imageUUID,
+	imageLookupFormat,
+	imageLookupBaseOS,
+	k8sVersion *string,
+) (string, error) {
+	if imageLookupFormat == nil {
+		imageLookupFormat = ptr.To("nkp-{{.BaseOS}}-.*?{{.K8sVersion}}-.*")
```
supershal, line 315: we can skip using `nkp` as it is internal acronym. Our CLI/UI can setup format, if this default is not good enough. 
```suggestion
		imageLookupFormat = ptr.To("*.{{.BaseOS}}-.*?{{.K8sVersion}}-.*")
```

We can also set it up in the API schema as default.

supershal:
We had briefly discussed to lookup images based on categories previously. However we didnt reached to final conclusion. Have we already consider that approach? I think @thunderboltsid had more context about complications using categories.

controllers/helpers.go
faiq: yeah, that's a good idea. this is totally the wrong place to do it :sweat_smile:

config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixmachines.yaml
faiq: i used the same convention as CAPA and other providers which has them on the top level.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/07a93a51edacdf97c326c8e5f3016b41be6bab40/api/v1beta2/awsmachine_types.go#L84

this is also used by tinkerbell 

https://github.com/tinkerbell/cluster-api-provider-tinkerbell/blob/main/api/v1beta1/tinkerbellmachine_types.go#L49

controllers/helpers.go
faiq: in our case, the name includes the timestamp but we can change this function to explicitly use the date.

config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixmachines.yaml
dlipovetsky: We should have a reason for diverging from the convention, and I think @supershal gives a good one.

(As an aside, that tinkerbell change was authored by a former CAPA maintainer 4 years ago.)

controllers/helpers.go
dlipovetsky:
```suggestion
		imageLookupFormat = ptr.To("capx-{{.BaseOS}}-.*?{{.K8sVersion}}-.*")
```

I think it might be safer to default to the `capx` prefix. WDYT?

templates/cluster-template.yaml
adiantum: Did you define BASE_OS and LOOKUP_FORMAT somewhere for E2E tests?

controllers/helpers.go
faiq: changed the default format to use this!

templates/cluster-template.yaml
faiq: i will start a new e2e test for this feature.

api/v1beta1/nutanixmachine_types.go
```diff
@@ -47,7 +47,30 @@ const (
...
+	// +kubebuilder:default:="{{.BaseOS}}-kube-{{.K8sVersion}}.qcow2"
+	Format *string `json:"format,omitempty"`
+	// ImageLookupBaseOS is the name of the base operating system to use for
+	// image lookup.
+	// +kubebuilder:validation:Required
+	BaseOS *string `json:"baseOS"`
+}
+
 // NutanixMachineSpec defines the desired state of NutanixMachine
+// +kubebuilder:validation:XValidation:rule="has(self.image) != has(self.imageLookup)",message="Either 'image' or 'imageLookup' must be set, but not both"
```
supershal, line 73: good to know about inline validation.

controllers/helpers.go
```diff
@@ -299,45 +302,136 @@ func GetSubnetUUID(ctx context.Context, client *prismclientv3.Client, peUUID str
...
-		// Validate filtered Images
-		foundImages := make([]*prismclientv3.ImageIntentResponse, 0)
-		for _, s := range responseImages.Entities {
-			imageSpec := s.Spec
-			if strings.EqualFold(*imageSpec.Name, *imageName) {
-				foundImages = append(foundImages, s)
-			}
+		return resolvedUUID, nil
+	}
+	return "", fmt.Errorf("failed to retrieve image by name or uuid. Verify input parameters")
```
supershal, line 339:
```suggestion
	return "", fmt.Errorf("failed to retrieve image by name or uuid or image lookup format. Verify input parameters")
```

dlipovetsky:
nit: I'd like to replace `K8sVersion` with `KubernetesVersion`, or even the more verbose (but more precise) `MachineKubernetesVersion`, e.g.

```
imageLookup:
  baseOS: rocky-9.5
  format: nkp-{{.BaseOS}}-.*?{{.MachineKubernetesVersion}}-.*
```

controllers/helpers.go
```diff
@@ -299,45 +302,136 @@ func GetSubnetUUID(ctx context.Context, client *prismclientv3.Client, peUUID str
...
 		if err != nil {
-			if strings.Contains(fmt.Sprint(err), "ENTITY_NOT_FOUND") {
-				return "", fmt.Errorf("failed to find image with UUID %s: %v", *imageUUID, err)
-			}
+			return "", err
 		}
-		foundImageUUID = *imageIntentResponse.Metadata.UUID
-	} else { // else search by name
-		responseImages, err := client.V3.ListAllImage(ctx, "")
+		return resolvedUUID, nil
```
dlipovetsky, line 330: Unfortunately, this function is replaced in https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/489. That is about to merge.

templates/image-lookup/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Nutanix. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+metadata:
+  name: nutanix-csi-kustomize
```
supershal, line 8: nit
```suggestion
  name: nutanix-image-lookup-kustomize
```

test/e2e/data/infrastructure-nutanix/v1beta1/cluster-template-image-lookup/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Nutanix. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+metadata:
+  name: nutanix-csi-kustomize
```
supershal, line 8: nit
```suggestion
  name: nutanix-image-lookup-kustomize
```

api/v1beta1/nutanixmachine_types.go
faiq: CEL is the best!

templates/image-lookup/lookup-patch.yaml
```diff
@@ -0,0 +1,24 @@
...
+      vcpusPerSocket: ${NUTANIX_MACHINE_VCPU_PER_SOCKET=1}
+      vcpuSockets: ${NUTANIX_MACHINE_VCPU_SOCKET=2}
+      memorySize: "${NUTANIX_MACHINE_MEMORY_SIZE=4Gi}"
+      systemDiskSize: "${NUTANIX_SYSTEMDISK_SIZE=40Gi}"
+      cluster:
+        type: name
+        name: "${NUTANIX_PRISM_ELEMENT_CLUSTER_NAME}"
+      subnet:
+        - type: name
+          name: "${NUTANIX_SUBNET_NAME}"
```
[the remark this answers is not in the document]
faiq: i suppose we don't and can add things via a couple of patches, but its just easier this way :D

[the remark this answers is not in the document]
faiq: lets just do it this lazy way -- i'd really like to resolve this PR as soon as we can.

api/v1beta1/nutanixmachine_types.go
```diff
@@ -47,7 +47,30 @@ const (
...
+type NutanixImageLookup struct {
+	// ImageLookupFormat is the naming format to look up the image for this
+	// machine It will be ignored if an explicit image is set. Supports
+	// substitutions for {{.BaseOS}} and {{.K8sVersion}} with the base OS and
+	// kubernetes version, respectively. The BaseOS will be the value in
+	// ImageLookupBaseOS and the kubernetes version as
+	// defined by the packages produced by kubernetes/release without v as a
+	// prefix: 1.13.0, 1.12.5-mybuild.1, or 1.17.3. For example, the default
+	// image format of {{.BaseOS}}-?{{.K8sVersion}}-* and ImageLookupBaseOS as "rhel-8.10" will end up
+	// searching for AMIs that match the pattern nkp-rhel-8.10-1.30.5-* for a
```
yannickstruyf3, line 61: AMIs are called Images in Nutanix context

controllers/helpers.go
faiq: rebased thanks for the heads up!

supershal (approved):
Thank you for nice feature improvement.

controllers/helpers.go
```diff
@@ -298,41 +302,136 @@ func GetSubnetUUID(ctx context.Context, client *prismclientv3.Client, peUUID str
...
+// GetImage returns an image UUID. If no UUID is provided, returns the unique image with the name.
 // Returns an error if no image has the UUID, if no image has the name, or more than one image has the name.
-func GetImage(ctx context.Context, client *prismclientv3.Client, id infrav1.NutanixResourceIdentifier) (*prismclientv3.ImageIntentResponse, error) {
-	switch {
-	case id.IsUUID():
-		resp, err := client.V3.GetImage(ctx, *id.UUID)
-		if err != nil {
-			if strings.Contains(fmt.Sprint(err), "ENTITY_NOT_FOUND") {
-				return nil, fmt.Errorf("failed to find image with UUID %s: %v", *id.UUID, err)
+func GetImage(
```
dlipovetsky, line 307: Please, to keep the code easy to understand and test, let `GetImage` remain a way to get an image from a resource identifier, i.e., a UUID or name). 

The caller can decide whether to call `getImageWithFormat`, e.g., if `spec.ImageLookup != nil`, or `GetImage` otherwise.

We can export `getImageWithFormat` , and perhaps rename it to `GetMostRecentImageWithFormat`, to make its semantics clear.

api/v1beta1/nutanixmachine_types.go
```diff
@@ -47,7 +47,30 @@ const (
...
+	// image format of {{.BaseOS}}-?{{.K8sVersion}}-* and ImageLookupBaseOS as "rhel-8.10" will end up
+	// searching for images that match the pattern nkp-rhel-8.10-1.30.5-* for a
+	// Machine that is targeting kubernetes v1.30.5. See
+	// also: https://golang.org/pkg/text/template/
+	// +kubebuilder:default:="{{.BaseOS}}-kube-v{{.K8sVersion}}.*"
+	Format *string `json:"format,omitempty"`
+	// ImageLookupBaseOS is the name of the base operating system to use for
+	// image lookup.
+	// +kubebuilder:validation:Required
+	BaseOS *string `json:"baseOS"`
```
dlipovetsky, line 69: Since the field is required, I think we don't want it to be a pointer.

api/v1beta1/nutanixmachine_types.go
```diff
@@ -144,14 +169,13 @@ type NutanixMachineStatus struct {
...
-//+kubebuilder:object:root=true
-//+kubebuilder:resource:path=nutanixmachines,shortName=nma,scope=Namespaced,categories=cluster-api
-//+kubebuilder:subresource:status
-//+kubebuilder:storageversion
-//+kubebuilder:printcolumn:name="Address",type="string",JSONPath=".status.addresses[0].address",description="The VM address"
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=nutanixmachines,shortName=nma,scope=Namespaced,categories=cluster-api
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="Address",type="string",JSONPath=".status.addresses[0].address",description="The VM address"
```
dlipovetsky, line 176: Did some tool (kubebuilder?) make this and other whitespace changes to this file? None of them are directly related to your change.

faiq: yeah i think this is kubebuilder that is doing this.

api/v1beta1/nutanixmachine_types.go
faiq: The validation becomes tricky here. With a pointer we are able to do validation on it not being nil, because it's a required value

```
// +kubebuilder:validation:MinLength
``` 

we can try that with a string.

I am picking up some pointers from this [blog post](https://ahmet.im/blog/crd-generation-pitfalls/index.html#explicit-required-or-optional-on-every-field) but i would like to hear your suggestions too.

controllers/helpers.go
faiq: Interesting, so move the responsibility of deciding which field to use back out to the caller. I think that makes sense.

faiq: Moved that logic out and left the function as is.

api/v1beta1/nutanixmachine_types.go
faiq: Changed the validation can you please take a look?

api/v1beta1/nutanixmachine_types.go
```diff
@@ -65,8 +65,8 @@ type NutanixImageLookup struct {
 	Format *string `json:"format,omitempty"`
 	// ImageLookupBaseOS is the name of the base operating system to use for
 	// image lookup.
-	// +kubebuilder:validation:Required
-	BaseOS *string `json:"baseOS"`
+	// +kubebuilder:validation:MinLength:=1
+	BaseOS string `json:"baseOS"`
```
thunderboltsid, line 69: In context of discussion yesterday, if we extend lookup with categories, this could field be optional too. so, let's not treat it like a required field

api/v1beta1/nutanixmachine_types.go
```diff
@@ -70,8 +93,13 @@ type NutanixMachineSpec struct {
 	// image is to identify the rhcos image uploaded to the Prism Central (PC)
 	// The image identifier (uuid or name) can be obtained from the Prism Central console
 	// or using the prism_central API.
-	// +kubebuilder:validation:Required
-	Image NutanixResourceIdentifier `json:"image"`
+	// +kubebuilder:validation:Optional
+	// +optional
+	Image *NutanixResourceIdentifier `json:"image,omitempty"`
```
deepakm-ntnx, line 98: we are going from required to now both optional. how are we conveying to user if they dont given any of it. can we  do something like following?
<pre>
image: <=== mandatory
      id *NutanixResourceIdentifier <==optional if imageLookup is provided
      imageLookup: <=== optional if id is provided
      ...
</pre>

api/v1beta1/nutanixmachine_types.go
```diff
@@ -47,7 +47,30 @@ const (
...
+	// ImageLookupBaseOS and the kubernetes version as
+	// defined by the packages produced by kubernetes/release without v as a
+	// prefix: 1.13.0, 1.12.5-mybuild.1, or 1.17.3. For example, the default
+	// image format of {{.BaseOS}}-?{{.K8sVersion}}-* and ImageLookupBaseOS as "rhel-8.10" will end up
+	// searching for images that match the pattern nkp-rhel-8.10-1.30.5-* for a
+	// Machine that is targeting kubernetes v1.30.5. See
+	// also: https://golang.org/pkg/text/template/
+	// +kubebuilder:default:="{{.BaseOS}}-kube-v{{.K8sVersion}}.*"
+	Format *string `json:"format,omitempty"`
+	// ImageLookupBaseOS is the name of the base operating system to use for
```
dlipovetsky, line 66:
```suggestion
	// BaseOS is the name of the base operating system to use for
```

api/v1beta1/nutanixmachine_types.go
```diff
@@ -47,7 +47,30 @@ const (
 	NutanixMachineBootstrapRefKindImage = "Image"
 )
 
+// NutanixImageLookup defines how to fetch images for the cluster
+// using the fields combined.
+type NutanixImageLookup struct {
+	// ImageLookupFormat is the naming format to look up the image for this
```
dlipovetsky, line 53:
```suggestion
	// Format is the naming format to look up the image for this
```

api/v1beta1/nutanixmachine_types.go
faiq: https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/498/files#diff-22f4b13037c2fbd14d0c08f995ed39a4460b298da33bc3d4b1d718f25da0f14bR73

this is handled by the CEL expression adding the line below in case that link doesn't work

```
// +kubebuilder:validation:XValidation:rule="has(self.image) != has(self.imageLookup)",message="Either 'image' or 'imageLookup' must be set, but not both"
``` 

if they are both missing it will be `false != false` which will fail.

api/v1beta1/nutanixmachine_types.go
faiq: This becomes tricky then, i think we should leave this as required for now and add it to be optional when we support category look up

controllers/helpers.go
```diff
@@ -301,41 +304,120 @@ func GetSubnetUUID(ctx context.Context, client *prismclientv3.Client, peUUID str
 	return foundSubnetUUID, nil
 }
 
-// GetImage returns an image. If no UUID is provided, returns the unique image with the name.
+// GetImage returns an image UUID. If no UUID is provided, returns the unique image with the name.
```
dlipovetsky, line 307: This function retursn an image, not a UUID.

controllers/helpers.go
```diff
@@ -301,41 +304,120 @@ func GetSubnetUUID(ctx context.Context, client *prismclientv3.Client, peUUID str
...
-		// Validate filtered Images
-		foundImages := make([]*prismclientv3.ImageIntentResponse, 0)
-		for _, s := range responseImages.Entities {
-			imageSpec := s.Spec
-			if strings.EqualFold(*imageSpec.Name, *id.Name) {
-				foundImages = append(foundImages, s)
-			}
+		return resolvedUUIDResponse, nil
+	default:
+	}
```
dlipovetsky, line 324: Unless this refactoring is required for the new functionality (and I don't think it is--neither `getImageByUUID` nor `getImageByName` are called by `GetImageWithFormat`), please remove it from this PR.

dlipovetsky: (I think that you were originally refactoring `GetImage` to support lookups, but now that you're moved that out to `GetImageWithFormat`, I think the refactoring is not necessary, and removing it would make it easier for everyone to review the PR)

faiq: good idea, removed.

controllers/helpers.go
faiq: fixed the comment

dlipovetsky:
Thanks a lot for removing the refactoring!

This looks good to me now.

Codecov is complaining about missing unit tests. I see that you did add e2e tests to exercise image lookup.

controllers/helpers.go
```diff
@@ -303,7 +306,7 @@ func GetSubnetUUID(ctx context.Context, client *prismclientv3.Client, peUUID str
 
 // GetImage returns an image. If no UUID is provided, returns the unique image with the name.
 // Returns an error if no image has the UUID, if no image has the name, or more than one image has the name.
-func GetImage(ctx context.Context, client *prismclientv3.Client, id infrav1.NutanixResourceIdentifier) (*prismclientv3.ImageIntentResponse, error) {
+func GetImage(ctx context.Context, client *prismclientv3.Client, id *infrav1.NutanixResourceIdentifier) (*prismclientv3.ImageIntentResponse, error) {
```
adiantum, line 309: What's the reason to change the param to pointer?

controllers/helpers.go
```diff
@@ -338,6 +341,64 @@ func GetImage(ctx context.Context, client *prismclientv3.Client, id infrav1.Nuta
...
+		return nil, fmt.Errorf(
+			"failed to substitute string %s with params %v error: %w",
+			*imageTemplate,
+			params,
+			err,
+		)
+	}
+	filterString := templateBytes.String()
+	filter := fmt.Sprintf("name==%s$", filterString)
+	responseImages, err := client.V3.ListAllImage(ctx, filter)
```
adiantum, line 376: Please do not use FIQL, it contains bugs. We're especially get rid from the FIQL in our projects.

controllers/helpers.go
```diff
@@ -338,6 +341,64 @@ func GetImage(ctx context.Context, client *prismclientv3.Client, id infrav1.Nuta
 	}
 }
 
+type ImageLookup struct {
+	BaseOS     string
+	K8sVersion string
+}
+
+func GetImageWithFormat(
```
adiantum, line 349: I really don't like naming here.

Try to think about GetImageByLookup or something similar which will connect the function with ImageLookup.

controllers/helpers.go
```diff
@@ -769,7 +830,8 @@ func getPrismCentralClientForCluster(ctx context.Context, cluster *infrav1.Nutan
 	managementEndpoint, err := clientHelper.BuildManagementEndpoint(ctx, cluster)
 	if err != nil {
 		log.Error(err, fmt.Sprintf("error occurred while getting management endpoint for cluster %q", cluster.GetNamespacedName()))
-		conditions.MarkFalse(cluster, infrav1.PrismCentralClientCondition, infrav1.PrismCentralClientInitializationFailed, capiv1.ConditionSeverityError, err.Error()) //nolint:govet // error will not be a constant
+		//nolint:govet // ignore error.
```
adiantum, line 833: "// error will not be a constant" better explain the reason for no lint

controllers/helpers.go
```diff
@@ -794,7 +856,9 @@ func getPrismCentralV4ClientForCluster(ctx context.Context, cluster *infrav1.Nut
 	managementEndpoint, err := clientHelper.BuildManagementEndpoint(ctx, cluster)
 	if err != nil {
 		log.Error(err, fmt.Sprintf("error occurred while getting management endpoint for cluster %q", cluster.GetNamespacedName()))
-		conditions.MarkFalse(cluster, infrav1.PrismCentralV4ClientCondition, infrav1.PrismCentralV4ClientInitializationFailed, capiv1.ConditionSeverityError, err.Error()) //nolint:govet // error will not be a constant
+		//nolint:govet // ignore error.
```
adiantum, line 859: "// error will not be a constant" better explain the reason for no lint

controllers/helpers.go
```diff
@@ -804,7 +868,9 @@ func getPrismCentralV4ClientForCluster(ctx context.Context, cluster *infrav1.Nut
 	})
 	if err != nil {
 		log.Error(err, "error occurred while getting nutanix prism v4 client from cache")
-		conditions.MarkFalse(cluster, infrav1.PrismCentralV4ClientCondition, infrav1.PrismCentralV4ClientInitializationFailed, capiv1.ConditionSeverityError, err.Error()) //nolint:govet // error will not be a constant
+		//nolint:govet // ignore error.
```
adiantum, line 871: "// error will not be a constant" better explain the reason for no lint

controllers/nutanixmachine_controller.go
```diff
@@ -775,7 +775,23 @@ func getDiskList(rctx *nctx.MachineContext) ([]*prismclientv3.VMDisk, error) {
...
+		)
+	} else if rctx.NutanixMachine.Spec.ImageLookup != nil {
+		nodeOSImage, err = GetImageWithFormat(
+			rctx.Context,
+			rctx.NutanixClient,
+			rctx.NutanixMachine.Spec.ImageLookup.Format,
+			&rctx.NutanixMachine.Spec.ImageLookup.BaseOS,
+			rctx.Machine.Spec.Version,
+		)
+	}
```
adiantum, line 794: What if both Image and ImageLookup are nil?

controllers/nutanixmachine_controller.go
```diff
@@ -803,7 +819,7 @@ func getSystemDisk(rctx *nctx.MachineContext) (*prismclientv3.VMDisk, error) {
 }
 
 func getBootstrapDisk(rctx *nctx.MachineContext) (*prismclientv3.VMDisk, error) {
-	bootstrapImageRef := infrav1.NutanixResourceIdentifier{
+	bootstrapImageRef := &infrav1.NutanixResourceIdentifier{
```
adiantum, line 822: Why do we need pointer here?

adiantum (changes requested):
Please take a look at the comments.

Regarding the image lookup feature: 
IMHO it would be great to use image lookup or image explicitly (some field to define it). Implicit implementation of this type features usually leads to misunderstanding and lack of supportability. That's not the thing that we would like for our customers.

controllers/nutanixmachine_controller.go
faiq: this is an invalid state because of the CEL statement.

controllers/nutanixmachine_controller.go
faiq: the function accepts a pointer.

controllers/helpers.go
faiq: Image is now a pointer because it's a pointer on the nutanixMachine object. It is an optional item a user can pass in.

controllers/helpers.go
faiq: fair enough.

faiq:
> IMHO it would be great to use image lookup or image explicitly (some field to define it). Implicit implementation of this type features usually leads to misunderstanding and lack of supportability. That's not the thing that we would like for our customers.

It is explicit and enforced by the CEL.

```
// +kubebuilder:validation:XValidation:rule="has(self.image) != has(self.imageLookup)",message="Either 'image' or 'imageLookup' must be set, but not both"
``` 

this will error if both are set or neither are set.

controllers/helpers.go
faiq: changed the function.

controllers/helpers.go
faiq: We use FIQL in several of our products in NKP and it has been working fine for us. There is an e2e test that exercises this functionality as well and it's proven to work.

thunderboltsid:
fixing broken build-container job in https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/505

controllers/helpers.go
faiq: These situations are entirely different. There will be three different ways to grab an image

1. Name directly
2. UUID directly
3. Image filtering

I could understand that previously we wouldn't want to depend on on filtering, but since there are 2 other options that don't depend on it users can make an informed decision based on their PC version.

dlipovetsky: Situation (2) doesn't require filtering, so that's not part of this conversation.

Situations (1) and (3) both require filtering, and we do client-side filtering in (1). Because we want a consistent implementation, and because FIQL is deprecated, we should therefore use client-side filtering for (2), unless there is a compelling reason to _not_ do so.

Do you see a compelling reason?

(Implementing the client-side filter for (2) seems straightforward.)

faiq: for the sake of consistency i went with client side filtering for 3.

.github/workflows/build-dev.yaml
```diff
@@ -81,6 +81,7 @@ jobs:
           - "clusterctl-upgrade"
           - "clusterclass"
           - "capx-feature-test"
+          - "image-templating"
```
thunderboltsid, line 84: We are already running this test as part of `capx-feature-test` suite. We shouldn't have to run this twice in CI.

test/e2e/template_test.go
```diff
@@ -0,0 +1,103 @@
+//go:build e2e
+
+/*
+Copyright 2022 Nutanix
```
thunderboltsid, line 4: nit: change year to 2025

test/e2e/data/infrastructure-nutanix/v1beta1/cluster-template-image-lookup/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2023 Nutanix. All rights reserved.
```
thunderboltsid, line 1: nit: change year to 2025

templates/image-lookup/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Nutanix. All rights reserved.
```
thunderboltsid, line 1: nit: change year to 2025

test/e2e/data/infrastructure-nutanix/v1beta1/cluster-template-image-lookup/kustomization.yaml
faiq: resolved

test/e2e/template_test.go
faiq: resolved

.github/workflows/build-dev.yaml
faiq: resolved

templates/image-lookup/kustomization.yaml
faiq: resolved.

api/v1beta1/nutanixmachine_types.go
```diff
@@ -47,7 +47,30 @@ const (
...
 )
 
+// NutanixImageLookup defines how to fetch images for the cluster
+// using the fields combined.
+type NutanixImageLookup struct {
+	// Format is the naming format to look up the image for this
+	// machine It will be ignored if an explicit image is set. Supports
+	// substitutions for {{.BaseOS}} and {{.K8sVersion}} with the base OS and
+	// kubernetes version, respectively. The BaseOS will be the value in
+	// ImageLookupBaseOS and the kubernetes version as
```
faiq, line 57:
```suggestion
	// BaseOS and the kubernetes version as
```

api/v1beta1/nutanixmachine_types.go
```diff
@@ -47,7 +47,30 @@ const (
...
+	// defined by the packages produced by kubernetes/release without v as a
+	// prefix: 1.13.0, 1.12.5-mybuild.1, or 1.17.3. For example, the default
+	// image format of {{.BaseOS}}-?{{.K8sVersion}}-* and ImageLookupBaseOS as "rhel-8.10" will end up
```
faiq, line 60:
```suggestion
	// image format of {{.BaseOS}}-?{{.K8sVersion}}-* and BaseOS as "rhel-8.10" will end up
```

api/v1beta1/nutanixmachine_types.go
dlipovetsky:
```suggestion
	// ImageLookupBaseOS and K8sVersion is the value in the Machine .spec.version, with the v prefix removed. This is effectively the version as
```

controllers/helpers.go
adiantum: Yeah, but we can check that Image is nil outside the function and call the function with `*id`. IMHO it's not necessary to change the function

dlipovetsky:
> ```yaml
> spec:
>   imageSelection: imageLookup
>   image: <image>
>   imageLookup:
>     format: pretty-image-{{.BaseOS}}-{{.K8Sversion}}.img
>     baseOS: <base OS>
> ```
> 
> Using something like `imageSelection` (or some other better naming) showing explicitly to user what kind of image selection strategy they want to use and give us ability to extend functionality in future. And last but not least: It's not required to much changes in this PR ;-)

I think I understand what you're proposing; I guess I'll call it the "explicitly specified strategy" pattern. I see the upsides. But there are some downsides, too.

1. The user must know the strategy names, i.e. the possible values of `imageLookup`.

2. The `image` and `imageLookup` fields may have unused values, and that can be confusing. A couple examples:
    
    a. The `image` value is not used.
    ```yaml
    spec:
      imageSelection: imageLookup
      image: capx-rhel-8.10-1.31.4
      imageLookup:
        format: capx-{{.BaseOS}}-{{.K8Sversion}}.img
        baseOS: ubuntu-24.04
    ```
    
    b. The  `imageLookup` field is not used.
    ```yaml
    spec:
      imageSelection: image
      image: capx-rhel-8.10-1.31.4
      imageLookup:
        format: capx-{{.BaseOS}}-{{.K8Sversion}}.img
        baseOS: ubuntu-24.04
    ```

adiantum:
> > ```yaml
> > spec:
> >   imageSelection: imageLookup
> >   image: <image>
> >   imageLookup:
> >     format: pretty-image-{{.BaseOS}}-{{.K8Sversion}}.img
> >     baseOS: <base OS>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Using something like `imageSelection` (or some other better naming) showing explicitly to user what kind of image selection strategy they want to use and give us ability to extend functionality in future. And last but not least: It's not required to much changes in this PR ;-)
> 
> I think I understand what you're proposing; I guess I'll call it the "explicitly specified strategy" pattern. I see the upsides. But there are some downsides, too.
> 
> 1. The user must know the strategy names, i.e. the possible values of the `imageSelection` field.
> 2. The `image` and `imageLookup` fields may have unused values, and that can be confusing. A couple examples:
>    a. The `image` value is not used.
>    ```yaml
>    spec:
>      imageSelection: imageLookup
>      image: capx-rhel-8.10-1.31.4
>      imageLookup:
>        format: capx-{{.BaseOS}}-{{.K8Sversion}}.img
>        baseOS: ubuntu-24.04
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    b. The  `imageLookup` field is not used.
>    ```yaml
>    spec:
>      imageSelection: image
>      image: capx-rhel-8.10-1.31.4
>      imageLookup:
>        format: capx-{{.BaseOS}}-{{.K8Sversion}}.img
>        baseOS: ubuntu-24.04
>    ```

Regarding downsides you've noted:
1. It's good that user should know the strategy name. We can have default strategy (e.g. image), but user should take a look into docs to select other strategies. When we don't have "explicitly specified strategy" user also need to read the docs, but with implicit strategy user can make something intuitively, what could leads to some errors in future, because user intuition could be different from our intuition.
2. From one side it's not good to have unnecessary info in manifest, from other side user can have this unnecessary info as comments. So eventually user will decide if they manifest will be clean or not.

IMHO upsides little bit prevail downsides.

faiq:
> Regarding downsides you've noted:
It's good that user should know the strategy name. We can have default strategy (e.g. image), but user should take a look into docs to select other strategies. When we don't have "explicitly specified strategy" user also need to read the docs, but with implicit strategy user can make something intuitively, what could leads to some errors in future, because user intuition could be different from our intuition.
From one side it's not good to have unnecessary info in manifest, from other side user can have this unnecessary info as comments. So eventually user will decide if they manifest will be clean or not.

> IMHO upsides little bit prevail downsides.

If the user needs to read the docs then the user can also see that setting both `spec.image` and `spec.imageLookup` is not allowed. 

This change to show strategy also adds unnecessary stutter 

```
spec:
  imageSelection: imageLookup
  image: capx-rhel-8.10-1.31.4
  imageLookup:
    format: capx-{{.BaseOS}}-{{.K8Sversion}}.img
    baseOS: ubuntu-24.04
 ```
 
 Its already obvious that I want to use imageLookup when I set it.

**Comment by faiq** [comment]
I will also note that the way I'm doing it is clearly laid out in [convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#unions) 

> Sometimes, at most one of a set of fields can be set. For example, the [volumes] field of a PodSpec has 17 different volume type-specific fields, such as nfs and iscsi. All fields in the set should be [Optional](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required).

When we look at [`VolumeSource`](https://pkg.go.dev/k8s.io/api/core/v1#VolumeSource) we see the same pattern  implemented here. All of the fields are marked optional and there is no "strategy" to decide which source to go with.

File: api/v1beta1/nutanixmachine_types.go [hunk 35]
```diff
@@ -47,7 +47,30 @@ const (
...
+	// +kubebuilder:default:="capx-{{.BaseOS}}-{{.K8sVersion}}-*"
+	Format *string `json:"format,omitempty"`
+	// BaseOS is the name of the base operating system to use for
+	// image lookup.
+	// +kubebuilder:validation:MinLength:=1
+	BaseOS string `json:"baseOS"`
+}
+
 // NutanixMachineSpec defines the desired state of NutanixMachine
+// +kubebuilder:validation:XValidation:rule="has(self.image) != has(self.imageLookup)",message="Either 'image' or 'imageLookup' must be set, but not both"
```
jimmidyson, line 73: This CEL expression is pretty opaque but it does work!

api/v1beta1/nutanixmachine_types.go
```diff
@@ -70,8 +93,13 @@ type NutanixMachineSpec struct {
 	// image is to identify the rhcos image uploaded to the Prism Central (PC)
```
jimmidyson, line 93: `rhcos` in this context meaning `Red Hat Enterprise Linux CoreOS`?
```suggestion
	// image is to identify the machine image uploaded to the Prism Central (PC)
```

api/v1beta1/nutanixmachine_types.go
```diff
@@ -70,8 +93,13 @@ type NutanixMachineSpec struct {
...
 	// The image identifier (uuid or name) can be obtained from the Prism Central console
 	// or using the prism_central API.
-	// +kubebuilder:validation:Required
-	Image NutanixResourceIdentifier `json:"image"`
+	// +kubebuilder:validation:Optional
+	// +optional
+	Image *NutanixResourceIdentifier `json:"image,omitempty"`
+	// imageLookup is a container that holds how to look up rhcos images for the cluster.
```
jimmidyson, line 99:
```suggestion
	// imageLookup is a container that holds how to look up the machine image for the cluster.
```

jimmidyson (approved):
Signing off on the API (haven't reviewed controller implementation). Looks good to me!

adiantum:
> I will also note that the way I'm doing it is clearly laid out in [convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#unions)
> 
> > Sometimes, at most one of a set of fields can be set. For example, the [volumes] field of a PodSpec has 17 different volume type-specific fields, such as nfs and iscsi. All fields in the set should be [Optional](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required).
> 
> When we look at [`VolumeSource`](https://pkg.go.dev/k8s.io/api/core/v1#VolumeSource) we see the same pattern implemented here. All of the fields are marked optional and there is no "strategy" to decide which source to go with.

The convention explicitly requires all fields to be optional and does not impose any requirements regarding the patterns we discussed. However, given the assertive approach you have taken in this and other reviews/discussions, I have decided to discontinue this discussion.

api/v1beta1/nutanixmachine_types.go
faiq: This is tested in main_test.go with several scenarios.

api/v1beta1/nutanixmachine_types.go
faiq: I copied and pasted what we previously used. I can change this.

faiq:
> The convention explicitly requires all fields to be optional and does not impose any requirements regarding the patterns we discussed. However, given the assertive approach you have taken in this and other reviews/discussions, I have decided to discontinue this discussion.

Actually, it exemplifies this perfectly. All of the options are considered optional -- however just like the behavior in this PR only one of these can be set. This is the code that enforces this behavior: 

https://github.com/kubernetes/kubernetes/blob/50ba48369d0c5e8da144440fb4b7644536bda345/pkg/apis/core/validation/validation.go#L509 

We can see this in action with this example: 

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: multi-volume-pod
spec:
  containers:
  - name: app-container
    image: nginx
    volumeMounts:
    - name: emptydir-volume
      mountPath: /mnt/emptydir
  volumes:
  - name: emptydir-volume
    emptyDir: {}
    configMap:
      name: example-config
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: example-config
data:
  app.properties: |
    key1=value1
    key2=value2
```

controllers/helpers.go
```diff
@@ -338,6 +342,70 @@ func GetImage(ctx context.Context, client *prismclientv3.Client, id infrav1.Nuta
...
+func GetImageByLookup(
+	ctx context.Context,
+	client *prismclientv3.Client,
+	imageTemplate,
+	imageLookupBaseOS,
+	k8sVersion *string,
+) (*prismclientv3.ImageIntentResponse, error) {
+	if strings.Contains(*k8sVersion, "v") {
+		k8sVersion = ptr.To(strings.Replace(*k8sVersion, "v", "", 1))
+	}
```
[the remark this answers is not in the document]
faiq: We should strip the `v` it's easy to add this back if needed -- we have downstream components that do not add a `v` in front of the version.

config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixmachinetemplates.yaml
```diff
@@ -191,7 +191,7 @@ spec:
                         type: array
                       image:
                         description: |-
-                          image is to identify the rhcos image uploaded to the Prism Central (PC)
+                          image is to identify the nutanix machine image uploaded to the Prism Central (PC)
```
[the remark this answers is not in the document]
faiq: we should follow up with making it consistent in another PR. this is a suggestion on top of another suggestion https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/498#discussion_r1971395674

jimmidyson (approved):
Signing off API as discussed today. Have not reviewed implementation.

controllers/helpers.go
faiq: thats true. it would be unexpected to have a pointer type in the API and then have functions that require you to deference it before passing them as input.

thunderboltsid (approved):

controllers/helpers.go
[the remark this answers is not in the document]
thunderboltsid: the choice of whether to use v or not in name can be handled in the template being used. for the internal version, we should strip v.

PR #502, opened 2025-02-20
build: Allow internal contributors to test workflows that use pull_request_target

dlipovetsky:
Allows internal contributors to test workflow changes. Workflow changes will run using pull_request event. If a workflow uses secrets and is run in an external PR, the workflow will not be able to read the secrets. Non-workflow changes will run using pull_request_target.

Fixes #

- We **cannot** test these changes in the context of the PR, because the workflows are from the `main` branch, as the result of being triggered by the `pull_request_target` event.
- However, I am able to test that the changes work as expected by opening two separate PRs, with this PR's branch as the target. 
   - #503 demonstrates that a PR that does not change files under `.github` (i.e., workflows) triggers actions using `pull_request_target`.
   - #504 demonstrates that a PR that changes files under `.github` (i.e., workflows) triggers actions using `pull_request`.

For example, when I pushed a commit that changed a workflow, the action was triggered by the event `pull_request`:
image

Later, when I pushed a commit that did not change a workflow, the action was triggered by the event `pull_request_target`: 
image

**Release note**:

```release-note

```

### DIFF
--- a/.github/workflows/build-dev.yaml
+++ b/.github/workflows/build-dev.yaml
@@ -6,7 +6,12 @@ on:
     branches:
       - main
       - 'release-*'
+  pull_request:
+    paths:
+    - '.github/**'
   pull_request_target:
+    paths-ignore:
+    - '.github/**'
 jobs:
   build-container:
     runs-on: ubuntu-latest
--- a/.github/workflows/synopsys.yaml
+++ b/.github/workflows/synopsys.yaml
@@ -1,10 +1,15 @@
 name: Black Duck Policy Check
 on:
-  pull_request_target:
   push:
     branches:
       - main
       - 'release-*'
+  pull_request:
+    paths:
+    - '.github/**'
+  pull_request_target:
+    paths-ignore:
+    - '.github/**'
 
 jobs:
   security:


fatz (approved):
@dlipovetsky LGTM. But can we also open a test PR against this branch itself to verify this is working like we expect it?

thunderboltsid:
Screenshot 2025-02-21 at 14 28 16
it seems we have duplicate test runs now?

dlipovetsky:
> it seems we have duplicate test runs now?

I can't explain the behavior in _this_ PR, because my changes in this PR are not even supposed to have any effect. However, in #503, there are no duplicate tests. I used two different commits in #503: one that changes a workflow, and one that does not. Either the pull_request, or the pull_request_target events triggered tests, never both. You can see that in the screenshots I attached above.

fatz:
My fault should read all details carefully. Sorry. 

Let's go ahead and get it merged

thunderboltsid (approved):

thunderboltsid:
> > it seems we have duplicate test runs now?
> 
> I can't explain the behavior in _this_ PR, because my changes in this PR are not even supposed to have any effect. However, in #503, there are no duplicate tests. I used two different commits (at different times) in #503: one that changes a workflow, and one that does not. Either the pull_request, or the pull_request_target events triggered tests, never both. You can see that in the screenshots I attached above.

Reasoned more through this behavior and I think I get it. When you change the workflow file itself (which lives under .github/**), you end up triggering both event types:
* pull_request_target:
This event always runs the workflow as it exists on the base branch (the target of the PR). In your base branch, the workflow file hasn’t been modified, so its paths-ignore rule (ignoring changes to .github/**) isn’t applied. As a result, the pull_request_target trigger fires as usual.
* pull_request:
The head branch (the PR) now includes changes to the workflow file. Because you’ve set a path filter that explicitly includes .github/**, this event is also triggered in the PR context.

So we shouldn't see this in future (and don't see this in your example PR).

thunderboltsid:
fixing broken build-container job in https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/505

dlipovetsky:
> > > it seems we have duplicate test runs now?
> > 
> > 
> > I can't explain the behavior in _this_ PR, because my changes in this PR are not even supposed to have any effect. However, in #503, there are no duplicate tests. I used two different commits (at different times) in #503: one that changes a workflow, and one that does not. Either the pull_request, or the pull_request_target events triggered tests, never both. You can see that in the screenshots I attached above.
> 
> Reasoned more through this behavior and I think I get it. When you change the workflow file itself (which lives under .github/**), you end up triggering both event types:
> 
> * pull_request_target:
>   This event always runs the workflow as it exists on the base branch (the target of the PR). In your base branch, the workflow file hasn’t been modified, so its paths-ignore rule (ignoring changes to .github/**) isn’t applied. As a result, the pull_request_target trigger fires as usual.
> * pull_request:
>   The head branch (the PR) now includes changes to the workflow file. Because you’ve set a path filter that explicitly includes .github/**, this event is also triggered in the PR context.
> 
> So we shouldn't see this in future (and don't see this in your example PR).

Thank you for thinking through that! Great explanation.

I think this implies that the head of the workflow (i.e., `on:`, which defines the triggers) always executes in the source branch. However, if the  `pull_request_target` trigger is executed, then the body of the workflow (i.e., `jobs:`) is read from the target branch.

This reminds me of fork/exec, and we all know how intuitive _that_ is  :joy:

PR #507, opened 2025-03-13
fix(ci): Update blackduck workflow to use newer action

thunderboltsid:
Older action has been deprecated and phased out.


fatz (approved):

manoj-nutanix (approved):

.github/workflows/blackduck.yaml
```diff
@@ -0,0 +1,73 @@
...
+            ${{ runner.os }}-go-
+
+      - name: Build Project
+        run: devbox run -- make build
+
+      - name: Set up Java 11
+        uses: actions/setup-java@v4
+        with:
+          java-version: "11"
+          distribution: "adopt"
```
dlipovetsky, line 46: This is a dependency?

.github/workflows/blackduck.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Setup go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
```
dlipovetsky, line 51: Above, we built CAPX within the devbox environment, and of course that means that environment has go installed. Why do we need this?

.github/workflows/blackduck.yaml
thunderboltsid: the scanner itself is java-based https://github.com/marketplace/actions/detect-rapid-scan-action#set-up-java

.github/workflows/blackduck.yaml
thunderboltsid: Good point, I had originally lifted the workflow spec from [CAREN blackduck workflow](https://github.com/nutanix-cloud-native/cluster-api-runtime-extensions-nutanix/blob/main/.github/workflows/blackduck.yaml). Let me try removing it to see if this still works of if our devbox setup causes some go env pollution preventing a clean run for the tool.

PR #510, opened 2025-04-22
fix(CVE-2025-22869): Update golang.org/x/crypto

thunderboltsid:
Peg golang.org/x/crypto to 0.35.0

CVE URL: https://avd.aquasec.com/nvd/cve-2025-22869


go.mod
```diff
@@ -178,8 +178,8 @@ require (
 )
 
 replace (
-	// CVE fixes for https://avd.aquasec.com/nvd/2024/cve-2024-45337
-	golang.org/x/crypto => golang.org/x/crypto v0.31.0
+	// CVE fixes for https://avd.aquasec.com/nvd/cve-2025-22869
+	golang.org/x/crypto => golang.org/x/crypto v0.35.0
```
jimmidyson, line 182: Why does this need to be a replace and not just in requires?

adiantum (approved):
lgtm

go.mod
thunderboltsid: it's an indirect dependency that's being pulled in.

jimmidyson: Can't we still update the dependency in require though? Using replace means we need to remember to bump it in future manually...

thunderboltsid: Sure.

jimmidyson (approved):

PR #514, opened 2025-05-06
fix(lint): Update golangci-lint to v2

thunderboltsid:
migrate golangci-lint config using built-in migrate tooling.


adiantum (approved):
approve tests

fatz (approved):

PR #518, opened 2025-05-06
fix(build): Add pull_request_target check_approvals guardrail on blackduck scan

thunderboltsid:
check_approvals now used across build-dev and blackduck workflows.


adiantum (approved):
lgtm

PR #523, opened 2025-06-02
Add support for NutanixFailureDomain

yanhua121:
- Add support for CRD/controller of NutanixFailureDomain
- NutanixCluster CRD/controller changes to support NutanixFailureDomain
- NutanixMachine CRD/status and controller changes to support NutanixFailureDomain

Did  manual integration testing and added unit test cases.

### DIFF
--- a/controllers/nutanixmachine_controller.go
+++ b/controllers/nutanixmachine_controller.go
@@ -510,15 +510,56 @@ func (r *NutanixMachineReconciler) reconcileNormal(rctx *nctx.MachineContext) (r
 	return reconcile.Result{}, nil
 }
 
+func (r *NutanixMachineReconciler) validateFailureDomainRef(rctx *nctx.MachineContext, fdName string) (*infrav1.NutanixFailureDomain, error) {
+	// Fetch the referent failure domain object
+	fdObj := &infrav1.NutanixFailureDomain{}
+	fdKey := client.ObjectKey{Name: fdName, Namespace: rctx.NutanixMachine.Namespace}
+	err := r.Get(rctx.Context, fdKey, fdObj)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch the referent failure domain object %q: %w", fdName, err)
+	}
+
+	// Validate the failure domain configuration
+	pe := fdObj.Spec.PrismElementCluster
+	peUUID, err := GetPEUUID(rctx.Context, rctx.NutanixClient, pe.Name, pe.UUID)
+	if err != nil {
+		return nil, err
+	}
+
+	subnets := fdObj.Spec.Subnets
+	_, err = GetSubnetUUIDList(rctx.Context, rctx.NutanixClient, subnets, peUUID)
+	if err != nil {
+		return nil, err
+	}
+
+	return fdObj, nil
+}
+
 func (r *NutanixMachineReconciler) validateMachineConfig(rctx *nctx.MachineContext) error {
-	if rctx.Machine.Spec.FailureDomain == nil {
-		if len(rctx.NutanixMachine.Spec.Subnets) == 0 {
-			return fmt.Errorf("atleast one subnet is needed to create the VM %s if no failure domain is set", rctx.NutanixMachine.Name)
+	log := ctrl.LoggerFrom(rctx.Context)
+
+	fdName := rctx.Machine.Spec.FailureDomain
+	if fdName != nil && *fdName != "" {
+		log.WithValues("failureDomain", *fdName)
+		fdObj, err := r.validateFailureDomainRef(rctx, *fdName)
+		if err != nil {
+			log.Error(err, "Failed to validate the failure domain")
+			return err
 		}
-		if (rctx.NutanixMachine.Spec.Cluster.Name == nil || *rctx.NutanixMachine.Spec.Cluster.Name == "") &&
-			(rctx.NutanixMachine.Spec.Cluster.UUID == nil || *rctx.NutanixMachine.Spec.Cluster.UUID == "") {
-			return fmt.Errorf("cluster name or uuid are required to create the VM %s if no failure domain is set", rctx.NutanixMachine.Name)
-		}
+
+		// Update the NutanixMachine machine config based on the failure domain spec
+		rctx.NutanixMachine.Spec.Cluster = fdObj.Spec.PrismElementCluster
+		rctx.NutanixMachine.Spec.Subnets = fdObj.Spec.Subnets
+		rctx.NutanixMachine.Status.FailureDomain = &fdObj.Name
+		log.Info(fmt.Sprintf("Updated the NutanixMachine %s machine config from the failure domain %s configuration.", rctx.NutanixMachine.Name, fdObj.Name))
+	}
+
+	if len(rctx.NutanixMachine.Spec.Subnets) == 0 {
+		return fmt.Errorf("at least one subnet is needed to create the VM %s", rctx.NutanixMachine.Name)
+	}
+	if (rctx.NutanixMachine.Spec.Cluster.Name == nil || *rctx.NutanixMachine.Spec.Cluster.Name == "") &&
+		(rctx.NutanixMachine.Spec.Cluster.UUID == nil || *rctx.NutanixMachine.Spec.Cluster.UUID == "") {
+		return fmt.Errorf("cluster name or uuid are required to create the VM %s", rctx.NutanixMachine.Name)
 	}
 
 	diskSize := rctx.NutanixMachine.Spec.SystemDiskSize
@@ -1118,40 +1159,15 @@ func (r *NutanixMachineReconciler) GetSubnetAndPEUUIDs(rctx *nctx.MachineContext
 	if rctx == nil {
 		return "", nil, fmt.Errorf("cannot create machine config if machine context is nil")
 	}
-	log := ctrl.LoggerFrom(rctx.Context)
-	if rctx.Machine.Spec.FailureDomain == nil || *rctx.Machine.Spec.FailureDomain == "" {
-		log.V(1).Info("no failure domain found on machine. Directly searching for Prism Element cluster")
-		if rctx.NutanixMachine.Spec.Cluster.Name == nil && rctx.NutanixMachine.Spec.Cluster.UUID == nil {
-			return "", nil, fmt.Errorf("cluster name or uuid must be passed if failure domain is not configured")
-		}
-		if len(rctx.NutanixMachine.Spec.Subnets) == 0 {
-			return "", nil, fmt.Errorf("subnets must be passed if failure domain is not configured")
-		}
-		peUUID, err := GetPEUUID(rctx.Context, rctx.NutanixClient, rctx.NutanixMachine.Spec.Cluster.Name, rctx.NutanixMachine.Spec.Cluster.UUID)
-		if err != nil {
-			return "", nil, err
-		}
-		subnetUUIDs, err := GetSubnetUUIDList(rctx.Context, rctx.NutanixClient, rctx.NutanixMachine.Spec.Subnets, peUUID)
-		if err != nil {
-			return "", nil, err
-		}
-		return peUUID, subnetUUIDs, nil
+
+	peUUID, err := GetPEUUID(rctx.Context, rctx.NutanixClient, rctx.NutanixMachine.Spec.Cluster.Name, rctx.NutanixMachine.Spec.Cluster.UUID)
+	if err != nil {
+		return "", nil, err
 	}
 
-	log.V(1).Info("failure domain config found. Ignoring cluster config on machine object (if any)")
-
-	failureDomainName := *rctx.Machine.Spec.FailureDomain
-	failureDomain, err := GetFailureDomain(failureDomainName, rctx.NutanixCluster)
+	subnetUUIDs, err := GetSubnetUUIDList(rctx.Context, rctx.NutanixClient, rctx.NutanixMachine.Spec.Subnets, peUUID)
 	if err != nil {
-		return "", nil, fmt.Errorf("failed to find failure domain %s", failureDomainName)
-	}
-	peUUID, err := GetPEUUID(rctx.Context, rctx.NutanixClient, failureDomain.Cluster.Name, failureDomain.Cluster.UUID)
-	if err != nil {
-		return "", nil, fmt.Errorf("failed to find prism element uuid for failure domain %s", failureDomainName)
-	}
-	subnetUUIDs, err := GetSubnetUUIDList(rctx.Context, rctx.NutanixClient, failureDomain.Subnets, peUUID)
-	if err != nil {
-		return "", nil, fmt.Errorf("failed to find subnet uuids for failure domain %s", failureDomainName)
+		return "", nil, err
 	}
 
 	return peUUID, subnetUUIDs, nil
--- a/api/v1beta1/nutanixmachine_types.go
+++ b/api/v1beta1/nutanixmachine_types.go
@@ -275,6 +275,10 @@ type NutanixMachineStatus struct {
 	// Will be set in case of failure of Machine instance
 	// +optional
 	FailureMessage *string `json:"failureMessage,omitempty"`
+
+	// failureDomain is the name of the failure domain where this Machine has been placed in.
+	// +optional
+	FailureDomain *string `json:"failureDomain,omitempty"`
 }
 
 // +kubebuilder:object:root=true
@@ -284,6 +288,7 @@ type NutanixMachineStatus struct {
 // +kubebuilder:printcolumn:name="Address",type="string",JSONPath=".status.addresses[0].address",description="The VM address"
 // +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.ready",description="NutanixMachine ready status"
 // +kubebuilder:printcolumn:name="ProviderID",type="string",JSONPath=".spec.providerID",description="NutanixMachine instance ID"
+// +kubebuilder:printcolumn:name="FailureDomain",type="string",JSONPath=".status.failureDomain",description="NutanixMachine FailureDomain"
 // NutanixMachine is the Schema for the nutanixmachines API
 type NutanixMachine struct {
 	metav1.TypeMeta   `json:",inline"`
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixmachines.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_nutanixmachines.yaml
@@ -31,6 +31,10 @@ spec:
       jsonPath: .spec.providerID
       name: ProviderID
       type: string
+    - description: NutanixMachine FailureDomain
+      jsonPath: .status.failureDomain
+      name: FailureDomain
+      type: string
     name: v1beta1
     schema:
       openAPIV3Schema:
@@ -485,6 +489,10 @@ spec:
                   - type
                   type: object
                 type: array
+              failureDomain:
+                description: failureDomain is the name of the failure domain where
+                  this Machine has been placed in.
+                type: string
               failureMessage:
                 description: Will be set in case of failure of Machine instance
                 type: string
--- a/controllers/nutanixmachine_controller_test.go
+++ b/controllers/nutanixmachine_controller_test.go
@@ -156,14 +156,14 @@ func TestNutanixMachineReconciler(t *testing.T) {
 				})
 				g.Expect(err).ToNot(HaveOccurred())
 			})
-			It("returns no error if valid machine config is passed with failure domain", func() {
+			It("returns error if invalid machine config is passed with reference to not-exist failure domain", func() {
 				machine.Spec.FailureDomain = &r
 				err := reconciler.validateMachineConfig(&nctx.MachineContext{
 					Context:        ctx,
 					NutanixMachine: ntnxMachine,
 					Machine:        machine,
 				})
-				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(err).To(HaveOccurred())
 			})
 		})
 
@@ -204,32 +204,6 @@ func TestNutanixMachineReconciler(t *testing.T) {
 					{
 						Type: infrav1.NutanixIdentifierName,
 						Name: &r,
-					},
-				}
-				_, _, err := reconciler.GetSubnetAndPEUUIDs(&nctx.MachineContext{
-					Context:        ctx,
-					NutanixMachine: ntnxMachine,
-					Machine:        machine,
-					NutanixCluster: ntnxCluster,
-				})
-				g.Expect(err).To(HaveOccurred())
-			})
-			It("should error if machine has failure domain and but it is missing on nutanixCluster object", func() {
-				machine.Spec.FailureDomain = &r
-
-				_, _, err := reconciler.GetSubnetAndPEUUIDs(&nctx.MachineContext{
-					Context:        ctx,
-					NutanixMachine: ntnxMachine,
-					Machine:        machine,
-					NutanixCluster: ntnxCluster,
-				})
-				g.Expect(err).To(HaveOccurred())
-			})
-			It("should error if machine and nutanixCluster have failure domain and but nutanixClient is nil", func() {
-				machine.Spec.FailureDomain = &r
-				ntnxCluster.Spec.FailureDomains = []infrav1.NutanixFailureDomain{
-					{
-						Name: r,
 					},
 				}
 				_, _, err := reconciler.GetSubnetAndPEUUIDs(&nctx.MachineContext{


api/v1beta1/nutanixfailuredomain_types.go
```diff
@@ -0,0 +1,95 @@
...
+type PrismElementFailureDomainSpec struct {
+	// prismElement is to identify the Prism Element in the Prism Central for the failure domain.
+	// +kubebuilder:validation:Required
+	PrismElement NutanixResourceIdentifier `json:"prismElement"`
+
+	// subnets holds a list of identifiers (one or more) of the PE cluster's network subnets
+	// for the Machine's VM to connect to.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	Subnets []NutanixResourceIdentifier `json:"subnets"`
```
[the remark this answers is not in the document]
yanhua121: Fixed

api/v1beta1/nutanixcluster_types.go
```diff
@@ -64,7 +64,30 @@ type NutanixClusterSpec struct {
...
+type NutanixFailureDomainReference struct {
+	// name of the referent failure domain.
+	// Only the local reference (from the same namespace) to the failure domain object is supported.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+
+	// controlPlane indicates if the referenced failure domain is suited for control-plane nodes for the cluster
+	// +kubebuilder:validation:Required
+	ControlPlane bool `json:"controlPlane"`
```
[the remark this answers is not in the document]
thunderboltsid: Cluster.status.failureDomains needs to have control plane flag on them as that's how kubeadmcontrolplane controller figures out which failure domain to place the CP machines on.

api/v1beta1/nutanixfailuredomain_types.go
```diff
@@ -0,0 +1,97 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+const (
+	// NutanixFailureDomainV2Kind represents the Kind of NutanixFailureDomainV2
+	NutanixFailureDomainV2Kind = "NutanixFailureDomainV2"
+
+	// NutanixFailureDomainV2Finalizer is the finalizer used by the NutanixFailureDomainV2 controller to block
+	// deletion of the NutanixFailureDomainV2 object if there are references to this object in other CRs.
```
faiq, line 29:
```suggestion
	// deletion of the NutanixFailureDomainV2 object if there are references to this object by other resources.
```

controllers/nutanixcluster_controller.go
```diff
@@ -95,10 +95,42 @@ func (r *NutanixClusterReconciler) SetupWithManager(ctx context.Context, mgr ctr
...
+			&infrav1.NutanixFailureDomainV2{},
+			handler.EnqueueRequestsFromMapFunc(
+				r.mapNutanixFailureDomainV2ToNutanixCluster(),
+			),
+		).
 		WithOptions(copts).
 		Complete(r)
 }
 
+func (r *NutanixClusterReconciler) mapNutanixFailureDomainV2ToNutanixCluster() handler.MapFunc {
```
faiq, line 108: Don't we want to only return clusters that are using the failure domain? 

Also do you think we should add a fieldIndexer here?

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +321,94 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
 		return nil
 	}
-	log.V(1).Info("Reconciling failure domains for cluster")
-	// If failure domains is nil on status object, first create empty slice
-	if rctx.NutanixCluster.Status.FailureDomains == nil {
-		rctx.NutanixCluster.Status.FailureDomains = make(capiv1.FailureDomains, 0)
+
+	// Not found failure domain names
+	notFoundFDs := []string{}
+	// Map of mis-configured failure domain name and validation error message
```
faiq, line 342:
```suggestion
	// Map of misconfigured failure domain name and validation error message
```

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +321,94 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
+	misConfigFDs := map[string]string{}
+
+	for _, fdRef := range rctx.NutanixCluster.Spec.ControlPlaneFailureDomains {
+		// Fetch the referent failure domain object
+		fdObj := &infrav1.NutanixFailureDomainV2{}
+		fdKey := client.ObjectKey{Name: fdRef.Name, Namespace: rctx.NutanixCluster.Namespace}
+		err := r.Client.Get(rctx.Context, fdKey, fdObj)
+		if err != nil {
+			log.Error(err, fmt.Sprintf("Failed to fetch the referent failure domain object %q.", fdRef.Name))
+			notFoundFDs = append(notFoundFDs, fdRef.Name)
```
faiq, line 352: this doesn't necessarily mean that the failure domain was not found. 

please check to see if the error is specifically is an `IsNotFound` error.

api/v1beta1/nutanixfailuredomain_types.go
```diff
@@ -0,0 +1,97 @@
...
+	NutanixFailureDomainV2Finalizer = "nutanixfailuredomainv2.infrastructure.cluster.x-k8s.io"
+)
+
+// PrismElementFailureDomainSpec defines Prism Element specific failure domain spec.
+type PrismElementFailureDomainSpec struct {
+	// prismElement is to identify the Prism Element in the Prism Central for the failure domain.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="prismElement is immutable once set"
```
faiq, line 36: please add a test for this CEL expression. I did something similar here https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/ebf8cb724171dabc0042c71eb06d04477e20cd1e/main_test.go#L503

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +321,94 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
 
 func (r *NutanixClusterReconciler) reconcileFailureDomains(rctx *nctx.ClusterContext) error {
 	log := ctrl.LoggerFrom(rctx.Context)
-	if len(rctx.NutanixCluster.Spec.FailureDomains) == 0 {
-		log.V(1).Info("no failure domains defined on cluster")
-		conditions.MarkTrue(rctx.NutanixCluster, infrav1.NoFailureDomainsReconciled)
+	log.Info("Reconciling failure domains for cluster")
+
+	// First clear the failure domain related conditions
```
faiq, line 326: why do we need to delete this condition?

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +321,94 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
+			errMsg = fmt.Sprintf("Not-found failure domain objects: %v", notFoundFDs)
+		}
+		if len(misConfigFDs) > 0 {
+			errMsg1 := fmt.Sprintf("Mis-configured failure domains: %v", misConfigFDs)
+			if errMsg == "" {
+				errMsg = errMsg1
+			} else {
+				errMsg += ", " + errMsg1
+			}
+		}
```
faiq, line 389: this code is confusing to read. please refactor it and please use more descriptive variables

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +321,94 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
+
+		conditions.MarkFalse(rctx.NutanixCluster, infrav1.FailureDomainsValidatedCondition,
+			infrav1.FailureDomainsMisConfiguredReason, capiv1.ConditionSeverityWarning, errMsg)
 	}
-	conditions.MarkTrue(rctx.NutanixCluster, infrav1.FailureDomainsReconciled)
+
+	return nil
+}
+
+func (r *NutanixClusterReconciler) validateFailureDomainSpec(rctx *nctx.ClusterContext, fd *infrav1.NutanixFailureDomainV2) error {
```
faiq, line 398: add a comment describing what the function is going to validate. please add a unit test as well.

api/v1beta1/nutanixfailuredomain_types.go
yanhua121: Thanks. Will add unit tests soon.

api/v1beta1/nutanixcluster_types.go
```diff
@@ -96,6 +117,7 @@ type NutanixClusterStatus struct {
 // +kubebuilder:storageversion
 // +kubebuilder:printcolumn:name="ControlplaneEndpoint",type="string",JSONPath=".spec.controlPlaneEndpoint.host",description="ControlplaneEndpoint"
 // +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=".status.ready",description="in ready status"
+// +kubebuilder:printcolumn:name="FailureDomains",type="string",JSONPath=".status.failureDomains",description="NutanixCluster FailureDomains"
```
thunderboltsid, line 120: this is a list of failure domains and not a string.

controllers/nutanixcluster_controller.go
yanhua121: The notFoundFDs include those the fetching call of the FD failed, no matter what the specific error is. Because to move to the next step of validating the FD configuration, the FD object is required to fetch without any error.

controllers/nutanixcluster_controller.go
yanhua121: changed

api/v1beta1/nutanixcluster_types.go
yanhua121: Yes, it is map type and shows like. I think it's useful for the users to see the cluster's configured/validated failureDomains without its maniests.

NAME                                                     CONTROLPLANEENDPOINT   READY   FAILUREDOMAINS
nutanixcluster.infrastructure.cluster.x-k8s.io/capx-y5   10.23.131.101          true    {"fd-t1":{"controlPlane":true},"fd-t11":{"controlPlane":true},"fd-t2":{"controlPlane":true},"fd-t22":{"controlPlane":true}}

controllers/nutanixcluster_controller.go
yanhua121: Add more comment to explain

controllers/nutanixcluster_controller.go
yanhua121: Done

controllers/nutanixcluster_controller.go
yanhua121: Reword the error message.

controllers/nutanixcluster_controller.go
yanhua121: Refactored.

controllers/nutanixcluster_controller.go
yanhua121: Fixed.

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +327,99 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
+	// Map of misconfigured failure domain name and validation error message
+	misConfigFDs := map[string]string{}
+
+	for _, fdRef := range rctx.NutanixCluster.Spec.ControlPlaneFailureDomains {
+		// Fetch the referent failure domain object
+		fdObj := &infrav1.NutanixFailureDomainV2{}
+		fdKey := client.ObjectKey{Name: fdRef.Name, Namespace: rctx.NutanixCluster.Namespace}
+		err := r.Client.Get(rctx.Context, fdKey, fdObj)
+		if err != nil {
+			log.Error(err, fmt.Sprintf("Failed to fetch the referent failure domain object %q.", fdRef.Name))
```
faiq, line 357: we need to make a distinction between different kinds of errors here. please use errors.Is

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +327,99 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
 	}
-	for _, fd := range rctx.NutanixCluster.Spec.FailureDomains {
-		rctx.NutanixCluster.Status.FailureDomains[fd.Name] = capiv1.FailureDomainSpec{ControlPlane: fd.ControlPlane}
+	if len(misConfigFDs) > 0 {
+		errMsg1 := fmt.Sprintf("Misconfigured failure domains: %v", misConfigFDs)
+		if errMsg == "" {
+			errMsg = errMsg1
+		} else {
+			errMsg += ", " + errMsg1
+		}
```
faiq, line 400: see errors.Join `https://pkg.go.dev/errors#Join`

faiq (dismissed):
I'd still like to see a unit test for the CEL validation

yanhua121:
> I'd still like to see a unit test for the CEL validation

@faiq Please see the CEL immutable fields validation unit-tests at nutanixfailuredomainv2_controller_test.go L122-145.

controllers/nutanixfailuredomainv2_controller_test.go
```diff
@@ -0,0 +1,172 @@
...
+		Context("Update an NutanixFailureDomainV2 Configuration", func() {
+			It("should not allow to update failure domain prism element configuration", func() {
+				// Create the NutanixFailureDomainV2 object and expect creation success
+				g.Expect(k8sClient.Create(ctx, fdObj)).To(Succeed())
+
+				// change the PrismElement configuration
+				fd2 := fdObj.DeepCopy()
+				fd2.Spec.PrismElement.PrismElement.Name = &rstr2
+				err := k8sClient.Update(ctx, fd2)
+				g.Expect(err).To(HaveOccurred())
```
faiq, line 131: lets check the contents of the error to see if they are as expected

yanhua121: Added the error message check.

controllers/nutanixcluster_controller.go
yanhua121: changed

controllers/nutanixmachine_controller.go
```diff
@@ -1118,40 +1159,15 @@ func (r *NutanixMachineReconciler) GetSubnetAndPEUUIDs(rctx *nctx.MachineContext
...
 	if err != nil {
-		return "", nil, fmt.Errorf("failed to find prism element uuid for failure domain %s", failureDomainName)
+		return "", nil, err
 	}
-	subnetUUIDs, err := GetSubnetUUIDList(rctx.Context, rctx.NutanixClient, failureDomain.Subnets, peUUID)
+
+	subnetUUIDs, err := GetSubnetUUIDList(rctx.Context, rctx.NutanixClient, rctx.NutanixMachine.Spec.Subnets, peUUID)
 	if err != nil {
-		return "", nil, fmt.Errorf("failed to find subnet uuids for failure domain %s", failureDomainName)
+		return "", nil, err
```
faiq, line 1170: Any reason we needed to remove wrapping the errors?

yanhua121: Here in this function, failureDomain is not involved any more. So we don't need to wrap the error message to include the failureDomainName.

api/v1beta1/conditions.go
```diff
@@ -24,15 +24,25 @@ const (
...
+	// NoFailureDomainsConfiguredCondition indicates no failure domains have been configured
+	NoFailureDomainsConfiguredCondition capiv1.ConditionType = "NoFailureDomainsConfigured"
+
+	// FailureDomainsValidatedCondition indicates whether the failure domains are configured correctly or not.
+	FailureDomainsValidatedCondition capiv1.ConditionType = "FailureDomainsValidated"
+
+	// FailureDomainsMisConfiguredReason (Severity=Warning) indicates that some of the failure domains
+	// are misconfigured.
+	FailureDomainsMisConfiguredReason = "FailureDomainsMisConfigured"
 )
```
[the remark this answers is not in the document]
yanhua121: changed.

api/v1beta1/nutanixfailuredomain_types.go
```diff
@@ -0,0 +1,97 @@
...
+	Status NutanixFailureDomainV2Status `json:"status,omitempty"`
+}
+
+// GetConditions returns the set of conditions for this object.
+func (nfd *NutanixFailureDomainV2) GetConditions() capiv1.Conditions {
+	return nfd.Status.Conditions
+}
+
+// SetConditions sets the conditions on this object.
+func (nm *NutanixFailureDomainV2) SetConditions(conditions capiv1.Conditions) {
```
thunderboltsid, line 82: keep pointer receivers consistent (i.e. nfd)

api/v1beta1/nutanixcluster_types.go
```diff
@@ -64,7 +64,26 @@ type NutanixClusterSpec struct {
 	// +listType=map
 	// +listMapKey=name
 	// +optional
+	//
+	// Deprecated: This field is replaced by the field controlPlaneFailureDomains and will be removed in the next apiVersion.
+	//
 	FailureDomains []NutanixFailureDomain `json:"failureDomains"`
```
thunderboltsid, line 70: Please add omitempty to the tag (i.e. json:"failureDomains,omitempty") so that an empty slice doesn’t produce [] in the manifest when no domains are configured.

api/v1beta1/nutanixcluster_types.go
```diff
@@ -107,6 +129,8 @@ type NutanixCluster struct {
 }
 
 // NutanixFailureDomain configures failure domain information for Nutanix.
+//
+// Deprecated: This type is replaced by the NutanixFailureDomainV2 CRD type and will be removed in the next apiVersion.
 type NutanixFailureDomain struct {
```
thunderboltsid, line 134: Rename the existing inline type used under NutanixClusterSpec.failureDomains from NutanixFailureDomain to NutanixFailureDomainConfig since Kubernetes CRDs are purely “JSON on-the-wire” and driven by field tags (not Go type names), renaming your Go struct has zero impact on the existing schema or stored objects: as long as you keep the same json:"failureDomains" tag. Kubebuilder and controller-gen look solely at those JSON tags to build the OpenAPI schema in your CRD.

api/v1beta1/nutanixfailuredomain_types.go
```diff
@@ -0,0 +1,97 @@
...
+	Conditions []capiv1.Condition `json:"conditions,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=nutanixfailuredomainv2s,shortName=nfd,scope=Namespaced,categories=cluster-api
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
+
+// NutanixFailureDomainV2 is the Schema for the NutanixFailureDomainV2 API.
+type NutanixFailureDomainV2 struct {
```
thunderboltsid, line 68: Rename the CRD from NutanixFailureDomainV2 to NutanixFailureDomain. API versioning belongs in the group/version (e.g. infrastructure.nutanix/v1beta1), not in the Kind name itself. That keeps our API consistent with Kubernetes best practices and avoids confusion when we introduce future versions (v1beta2, etc.).

api/v1beta1/nutanixfailuredomain_types.go
yanhua121: changed

api/v1beta1/nutanixcluster_types.go
yanhua121: added

api/v1beta1/nutanixcluster_types.go
```diff
@@ -64,7 +64,26 @@ type NutanixClusterSpec struct {
...
+	ControlPlaneFailureDomains []NutanixFailureDomainReference `json:"controlPlaneFailureDomains,omitempty"`
+}
+
+// NutanixFailureDomainReference identifies the nutanix failure domain resource.
+type NutanixFailureDomainReference struct {
+	// name of the referent failure domain.
+	// Only the local reference (from the same namespace) to the failure domain object is supported.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
yanhua121: Removed

api/v1beta1/nutanixfailuredomain_types.go
yanhua121: Renamed to NutanixFailureDomain

api/v1beta1/nutanixcluster_types.go
```diff
@@ -64,7 +64,26 @@ type NutanixClusterSpec struct {
 	// +listType=map
 	// +listMapKey=name
 	// +optional
+	//
+	// Deprecated: This field is replaced by the field controlPlaneFailureDomains and will be removed in the next apiVersion.
+	//
```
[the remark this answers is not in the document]
yanhua121: It seems we cannot add the CEL rules to the optional/"omitempty" fields. When the CEL rule added to the NutanixClusterSpec, the nutanixcluster object creation (with empty of either of these fields in the spec) unit-tests all failed with the below error:

Create NutanixCluster object error: NutanixCluster.infrastructure.cluster.x-k8s.io "test" is invalid: spec: Invalid value: "object": no such key: controlPlaneFailureDomains evaluating rule: controlPlaneFailureDomains and failureDomains (deprecated) cannot be set simultaneously

The above test failures were with the below CEL rule added to the NutanixClusterSpec.

// +kubebuilder:validation:XValidation:rule="size(self.controlPlaneFailureDomains) == 0 || size(self.failureDomains) == 0",message="controlPlaneFailureDomains and failureDomains (deprecated) cannot be set simultaneously"

config/manager/kustomization.yaml
```diff
@@ -13,5 +13,5 @@ kind: Kustomization
 
 images:
 - name: controller
-  newName: ghcr.io/nutanix-cloud-native/cluster-api-provider-nutanix/controller
-  newTag: latest
+  newName: ko.local/cluster-api-provider-nutanix
+  newTag: e2e-432658dd426691c79d167a7502036904f19d87d2
```
thunderboltsid, line 17: this is pointing GitHub container registry, when we create releases it uses these as manifests so let's not change this. Instead, set env var LOCAL_IMAGE_REGISTRY to your internal OCI registry before running tests instead.

yanhua121: Thanks for helping catch this out. I checked in this by accident.

Fixed.

api/v1beta1/nutanixcluster_types.go
```diff
@@ -64,7 +64,17 @@ type NutanixClusterSpec struct {
...
-	FailureDomains []NutanixFailureDomain `json:"failureDomains"`
+	//
+	// Deprecated: This field is replaced by the field controlPlaneFailureDomains and will be removed in the next apiVersion.
+	//
+	FailureDomains []NutanixFailureDomainConfig `json:"failureDomains,omitempty"`
+
+	// controlPlaneFailureDomains configures references to the NutanixFailureDomain objects
+	// that the cluster uses to deploy its control-plane machines.
+	// +listType=map
+	// +listMapKey=name
```
faiq, line 75: I dont think we need these any more. local object reference is only names

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +328,81 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
+		fdObj := &infrav1.NutanixFailureDomain{}
+		fdKey := client.ObjectKey{Name: fdRef.Name, Namespace: rctx.NutanixCluster.Namespace}
+		err := r.Client.Get(rctx.Context, fdKey, fdObj)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				validationErrs = append(validationErrs, fmt.Errorf("not found the failure domain object with name %q", fdRef.Name))
+			} else {
+				validationErrs = append(validationErrs, fmt.Errorf("failed to fetch the failure domain object with name %q: %w", fdRef.Name, err))
+			}
+			continue
```
faiq, line 359: you can put this continue in

```
if errors.IsNotFound(err) {
				validationErrs = append(validationErrs, fmt.Errorf("not found the failure domain object with name %q", fdRef.Name))
				continue
			}
			
```

and remove the else

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +328,81 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
+		failureDomains[fd.Name] = capiv1.FailureDomainSpec{ControlPlane: fd.ControlPlane}
 	}
-	for _, fd := range rctx.NutanixCluster.Spec.FailureDomains {
-		rctx.NutanixCluster.Status.FailureDomains[fd.Name] = capiv1.FailureDomainSpec{ControlPlane: fd.ControlPlane}
+
+	// Set the failure domains for nutanixcluster status
+	rctx.NutanixCluster.Status.FailureDomains = failureDomains
+
+	if len(validationErrs) == 0 {
+		conditions.MarkTrue(rctx.NutanixCluster, infrav1.FailureDomainsValidatedCondition)
```
faiq, line 382: if we return here, we don't need an else.

api/v1beta1/nutanixcluster_types.go
yanhua121: Using the map type with key of "name" can make sure there is no duplicate failureDomain configuration in the list.

controllers/nutanixcluster_controller.go
```diff
@@ -19,6 +19,7 @@ package controllers
 import (
 	"cmp"
 	"context"
+	errs "errors"
```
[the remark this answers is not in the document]
yanhua121: This is to differentiate "k8s.io/apimachinery/pkg/api/errors" on L28.

controllers/nutanixcluster_controller.go
yanhua121: This if/else block each situation just has one statement, it won't affect readability.

yanhua121: Since this is close to the end of the function, it is not a big difference.

controllers/nutanixcluster_controller.go
faiq: actually never mind. i was wrong on this

api/v1beta1/nutanixcluster_types.go
faiq: if you want no duplicate names use this

```
// +kubebuilder:validation:UniqueItems
```
https://book.kubebuilder.io/reference/markers/crd-validation

controllers/nutanixcluster_controller.go
yanhua121: changed as suggested

controllers/nutanixcluster_controller_test.go
```diff
@@ -161,10 +143,8 @@ func TestNutanixClusterReconciler(t *testing.T) {
 				g.Expect(result.RequeueAfter).To(BeZero())
 				g.Expect(result.Requeue).To(BeFalse())
 			})
-			It("should not error and not requeue if failure domains are configured and cluster is Ready", func() {
-				ntnxCluster.Spec.FailureDomains = []infrav1.NutanixFailureDomain{
-					fd1,
-				}
+			It("should not error and not requeue if controlPlaneFailureDomains are configured and cluster is Ready", func() {
+				ntnxCluster.Spec.ControlPlaneFailureDomains = []corev1.LocalObjectReference{fdRef}
```
thunderboltsid, line 147: we should add a separate test for controlplanefailuredomains instead of modifying existing ones

controllers/nutanixcluster_controller.go
```diff
@@ -289,20 +328,81 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
+				validationErrs = append(validationErrs, fmt.Errorf("not found the failure domain object with name %q", fdRef.Name))
+				continue
+			}
+			validationErrs = append(validationErrs, fmt.Errorf("failed to fetch the failure domain object with name %q: %w", fdRef.Name, err))
+			continue
+		}
+
+		// Validate the failure domain configuration
+		err = r.validateFailureDomainSpec(rctx, fdObj)
+		if err != nil {
```
thunderboltsid, line 364: nit: limit the scope of err to nil check conditional

controllers/nutanixfailuredomain_controller.go
```diff
@@ -0,0 +1,235 @@
...
+func (r *NutanixFailureDomainReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, reterr error) {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Reconciling the NutanixFailureDomain")
+
+	var err error
+
+	// Fetch the NutanixFailureDomain instance
+	fd := &infrav1.NutanixFailureDomain{}
+	err = r.Get(ctx, req.NamespacedName, fd)
+	if err != nil {
```
thunderboltsid, line 141: nit: declare err in conditional scope

controllers/nutanixfailuredomain_controller.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// Handle deletion
+	if !fd.DeletionTimestamp.IsZero() {
+		return r.reconcileDelete(ctx, fd)
+	}
+
+	return r.reconcileNormal(ctx, fd)
+}
+
+func (r *NutanixFailureDomainReconciler) reconcileDelete(ctx context.Context, fd *infrav1.NutanixFailureDomain) (res ctrl.Result, reterr error) {
```
thunderboltsid, line 187: throughout function we have a explicit return, hence we don't need named return values. you can change signature to 
```go
reconcileDelete(ctx context.Context, fd *infrav1.NutanixFailureDomain) (ctrl.Result, error)
```

api/v1beta1/nutanixfailuredomain_types.go
```diff
@@ -0,0 +1,97 @@
...
+	capiv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+const (
+	// NutanixFailureDomainKind represents the Kind of NutanixFailureDomain
+	NutanixFailureDomainKind = "NutanixFailureDomain"
+
+	// NutanixFailureDomainFinalizer is the finalizer used by the NutanixFailureDomain controller to block
+	// deletion of the NutanixFailureDomain object if there are references to this object by other resources.
+	NutanixFailureDomainFinalizer = "nutanixfailuredomain.infrastructure.cluster.x-k8s.io"
```
thunderboltsid, line 30: nit: we should consider using a domain qualified finalizer e.g `infrastructure.cluster.x-k8s.io/nutanixfailuredomain` instead
see: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#finalizers

controllers/nutanixfailuredomain_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	errMsg := fmt.Sprintf("The failure domain is used by machines: %v", ntxMachines)
+	conditions.MarkFalse(fd, infrav1.FailureDomainSafeForDeletionCondition,
+		infrav1.FailureDomainInUseReason, capiv1.ConditionSeverityError, errMsg)
+
+	reterr = fmt.Errorf("the failure domain %q is not safe for deletion since it is in use", fd.Name)
+	log.Error(reterr, errMsg)
+	return ctrl.Result{}, reterr
+}
+
+func (r *NutanixFailureDomainReconciler) reconcileNormal(ctx context.Context, fd *infrav1.NutanixFailureDomain) (res ctrl.Result, reterr error) {
```
thunderboltsid, line 227: we have a explicit return, hence we don't need named return values.

controllers/nutanixfailuredomain_controller.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// Initialize the patch helper.
+	patchHelper, err := patch.NewHelper(fd, r.Client)
+	if err != nil {
+		log.Error(err, "Failed to configure the patch helper")
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	defer func() {
+		// Always attempt to Patch the NutanixCluster object and its status after each reconciliation.
```
thunderboltsid, line 161: nit: I'm assuming you mean the NutanixFailureDomain object.

api/v1beta1/nutanixcluster_types.go
faiq: i created a PR against this https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/531 to show that it does work. I prefer that we use this simpler validation

controllers/nutanixfailuredomain_controller.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Handle deletion
+	if !fd.DeletionTimestamp.IsZero() {
+		return r.reconcileDelete(ctx, fd)
+	}
+
+	return r.reconcileNormal(ctx, fd)
+}
+
+func (r *NutanixFailureDomainReconciler) reconcileDelete(ctx context.Context, fd *infrav1.NutanixFailureDomain) (ctrl.Result, error) {
```
thunderboltsid, line 183: Currently the reconcileDelete only looks at whether existing NutanixMachines are placed on a failure domain. If not, it removes the finalizer. In theory, this behavior should also extend to checking failure domains referenced in NutanixCluster.spec.ControlPlaneFailureDomains.

At the moment, the NutanixClusterController would remove the failure domain from the cluster's status to ensure a new control plane machine doesn't get created with a deleted failure domain and mark the NutanixCluster with a `FailureDomainsMisconfigured` condition.

We discussed offline on this a bit and decided not to hold the PR on this issue. However, I will experiment with some more scenarios and try to work out whether or not this should be the desired behavior.

thunderboltsid (approved):

PR #530, opened 2025-06-13
fix(build): Enable Test Execution When Labels Are Added to Existing PRs

rohithkumar-nutanix:
Fix for: Tests are not running if a label is added after a PR is created

### DIFF
--- a/.github/workflows/blackduck.yaml
+++ b/.github/workflows/blackduck.yaml
@@ -7,7 +7,9 @@ on:
       - main
       - 'release-*'
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
   pull_request_target:
+    types: [opened, synchronize, reopened, labeled]
     branches:
       - main
       - 'release-*'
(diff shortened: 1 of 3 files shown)


.github/workflows/blackduck.yaml
```diff
@@ -7,7 +7,9 @@ on:
       - main
       - 'release-*'
   pull_request:
+    types: [opened, synchronize, reopened, labeled]
```
thunderboltsid, line 10: my understanding is when the types field is not specified for a GitHub Actions trigger like pull_request or pull_request_target, the workflow will be triggered for all available activity types for that event. Hence, adding types explicitly (e.g. defining a subset of events) does not solve the issue of jobs not running if a label is added after the PR is created

rohithkumar-nutanix: By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened.
It is mentioned in the docs, it does not support labelled by default.
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows

thunderboltsid (approved):

mhrabovcin (approved):

manoj-nutanix (approved):

PR #546, opened 2025-07-18
fix(api): Add omitzero and omitempty to optional fields 

thunderboltsid:
- Cluster is an optional field of type struct: omitzero
- Subnets is a container type: omitempty

Note: omitzero semantics were introduced in Go1.24 hence go bump


dkoshkin (approved):

thunderboltsid:
Had to make some changes as the 5th parameter in `conditions.MarkFalse` https://pkg.go.dev/sigs.k8s.io/cluster-api/util/conditions#MarkFalse is actually a format string and the 6th parameter onwards, it's the value.

dlipovetsky (approved):

PR #549, opened 2025-07-23
NCN-108631:  Enforce CEL validation on NutanixResourceIdentifier

thunderboltsid:
only one of name and uuid should be specified depending on the
type chosen. UUIDs should be of a UUID format.


thunderboltsid (commented):
lgtm

atulv7 (approved):

api/v1beta1/nutanix_types.go
```diff
@@ -61,6 +61,8 @@ const (
 
 // NutanixResourceIdentifier holds the identity of a Nutanix PC resource (cluster, image, subnet, etc.)
 // +union
+// +kubebuilder:validation:XValidation:rule="self.type == 'name' ? (has(self.name) && size(self.name) > 0) : !has(self.name)",message="'name' must be set and non-empty when type is 'name', and forbidden otherwise"
```
jimmidyson, line 64: Min length of 1 for name can be added via kubebuilder validation annotation on `Name` property.

UUID pattern can be added via kubebuilder validation annotation on UUID property.

I would also add multiple CEL rules with more specific messages that just deals with the semantics

```suggestion
// +kubebuilder:validation:XValidation:rule="self.type != 'name' || has(self.name)",message="the 'name' field must be present when the 'type' is 'name'"
// +kubebuilder:validation:XValidation:rule="self.type != 'uuid' || has(self.uuid)",message="'the 'uuid' field must be present when the 'type' is 'uuid'"
// +kubebuilder:validation:XValidation:rule="self.type != 'name' || !has(self.uuid)",message="the 'uuid' field cannot be set when the 'type' is 'name'"
// +kubebuilder:validation:XValidation:rule="self.type == 'uuid' && !has(self.name)",message="the 'name' field cannot be set when the 'type' is 'uuid'"
```

api/v1beta1/nutanix_types.go
```diff
@@ -69,6 +71,8 @@ type NutanixResourceIdentifier struct {
 
 	// uuid is the UUID of the resource in the PC.
 	// +optional
+	// +kubebuilder:validation:MaxLength=36
+	// +kubebuilder:validation:MinLength=36
```
jimmidyson, line 75: We can use uuid format here - see https://github.com/kubernetes/apiextensions-apiserver/blob/v0.33.3/pkg/apiserver/validation/formats.go#L45-L48 for options and pick most appropriate one.

```suggestion
	// +kubebuilder:validation:Format=`uuid`
```

yanhua121 (commented):
@abhay-nutanix 
Move the unit tests to the controllers/ folder so you can use the existing k8sClient with the envtest there. Or you need to setup the envtest with CRD schema similar to that in the controllers/suite_test.go in the api/v1beta1 folder.

thunderboltsid:
> @abhay-nutanix Move the unit tests to the controllers/ folder so you can use the existing k8sClient with the envtest there. Or you need to setup the envtest with CRD schema similar to that in the controllers/suite_test.go in the api/v1beta1 folder.

isn't that already the case, https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/549/files#diff-7bb8f4003a2557fd11af8a3aa43d5b442c96358fdb2e1853cc1f067a828724a2R49-R71, or do you mean something else?

api/v1beta1/nutanix_types.go
abhay-nutanix: // name is the resource name in the PC
	// +optional
	// +kubebuilder:validation:MinLength=1
	Name *string `json:"name,omitempty"`
	
this is already added above the structure declaration

abhay-nutanix: incorporated the changes

api/v1beta1/nutanix_resourceidentifier_validation_test.go
```diff
@@ -0,0 +1,299 @@
...
+			machine := createTestNutanixMachine("test-invalid-uuid-no-dashes", infrav1.NutanixResourceIdentifier{
+				Type: infrav1.NutanixIdentifierUUID,
+				UUID: &invalidUUID,
+			})
+
+			err := k8sClient.Create(ctx, machine)
+			Expect(err).To(HaveOccurred())
+		})
+
+		It("should accept UUID with wrong dash positions (limited validation)", func() {
```
thunderboltsid, line 167: this should be changed to "should reject"

api/v1beta1/nutanix_resourceidentifier_validation_test.go
```diff
@@ -0,0 +1,299 @@
...
+				UUID: &invalidUUID,
+			})
+
+			err := k8sClient.Create(ctx, machine)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("uuid"))
+		})
+
+		It("should reject 36-char string without dashes", func() {
+			invalidUUID := "abcdefghijklmnopqrstuvwxyz1234567890" // 36 chars, no dashes
```
thunderboltsid, line 288: this would fail even with dashes because it's not hexadecimal chars

api/v1beta1/nutanix_resourceidentifier_validation_test.go
```diff
@@ -0,0 +1,299 @@
...
+				UUID: &validUUID, // This should not be allowed
+			})
+
+			err := k8sClient.Create(ctx, machine)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("uuid"))
+		})
+	})
+
+	Context("Edge cases and comprehensive validation", func() {
```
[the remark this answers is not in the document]
abhay-nutanix: Will remove this Context

api/v1beta1/nutanix_types.go
```diff
@@ -61,6 +61,8 @@ const (
 
 // NutanixResourceIdentifier holds the identity of a Nutanix PC resource (cluster, image, subnet, etc.)
 // +union
+// +kubebuilder:validation:XValidation:rule="self.type == 'name' ? has(self.name) : !has(self.name)",message="'name' must be set when type is 'name', and forbidden otherwise"
+// +kubebuilder:validation:XValidation:rule="self.type == 'uuid' ? has(self.uuid) && self.uuid.contains('-') : !has(self.uuid)",message="'uuid' configuration is required when type is 'uuid', and forbidden otherwise"
```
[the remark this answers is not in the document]
thunderboltsid: `"<field>" must be set when type is "<type>", and must not be set otherwise`

thunderboltsid:
Thanks a ton for shepherding through 🙌. The CEL side of things ended up being trickier than expected because of cost implications, especially with rules that touch unbounded lists/maps and trigger the budget estimator.

As a follow-up, we should spin a separate PR in the future that explicitly sets sensible maxItems, maxProperties, and maxLength on all containers where we rely on CEL rules. That way, we’ll keep validation expressive while staying well under the CEL budget ceiling.

PR #550, opened 2025-07-23
fix: check/update NutanixMachine.status.failureDomain properly in reconciler

yanhua121:
https://jira.nutanix.com/browse/NCN-108624
fix: make sure to check/update NutanixMachine.status.failureDomain properly in the nutanixmachine reconciler.

Currently the nutanixmachine reconciler only set the .status.failureDomain when the machine's VM is newly created with a FD, but the reconciler does not check and update this .status.failureDomain in each reconcile call. The fix is to check and update it properly for each reconcile call, similar to that of status.addresses.


controllers/nutanixmachine_controller.go
```diff
@@ -514,6 +521,33 @@ func (r *NutanixMachineReconciler) reconcileNormal(rctx *nctx.MachineContext) (r
...
+	fdKey := client.ObjectKey{Name: fdName, Namespace: rctx.NutanixMachine.Namespace}
+	err := r.Get(rctx.Context, fdKey, fdObj)
+	if err != nil {
+		return fmt.Errorf("failed to fetch the referent failure domain object %q: %w", fdName, err)
+	}
+
+	// Validate the NutanixMachine machine spec is consistent with that in the failure domain spec
+	if !reflect.DeepEqual(rctx.NutanixMachine.Spec.Cluster, fdObj.Spec.PrismElementCluster) ||
+		!reflect.DeepEqual(rctx.NutanixMachine.Spec.Subnets, fdObj.Spec.Subnets) {
+		return fmt.Errorf("The nutanixmachine's spec cluster/subnets is not consistent with that in the failure domain %q", *rctx.Machine.Spec.FailureDomain)
```
dlipovetsky, line 542: Did you consider comparing the fields individually, and reporting a detailed diff to the user? I feel that reporting "is not consistent" may be too little information.

Also, did you consider implementing this comparison in an `Equal` receiver of `NutanixResourceIdentifier`?

dlipovetsky:
Did you consider implementing an `EqualTo(other)` receiver (or something similar) on `NutanixResourceIdentifier`? It has a few receivers already.

controllers/nutanixmachine_controller.go
```diff
@@ -514,6 +521,33 @@ func (r *NutanixMachineReconciler) reconcileNormal(rctx *nctx.MachineContext) (r
...
+func (r *NutanixMachineReconciler) checkFailureDomainStatus(rctx *nctx.MachineContext) error {
+	if rctx.Machine.Spec.FailureDomain == nil || *rctx.Machine.Spec.FailureDomain == "" {
+		return nil
+	}
+
+	// Fetch the referent failure domain object
+	fdObj := &infrav1.NutanixFailureDomain{}
+	fdName := *rctx.Machine.Spec.FailureDomain
+	fdKey := client.ObjectKey{Name: fdName, Namespace: rctx.NutanixMachine.Namespace}
+	err := r.Get(rctx.Context, fdKey, fdObj)
```
dkoshkin, line 534: We already have this code inside the `validateMachineConfig` function, where the new FD changes are mutating the NutanixMachine.
https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/cd2407f11b24a785ce6df1f67646c001b939beb3/controllers/nutanixmachine_controller.go#L557

What do you think about combining the 2 in a new function like `reconcileFailureDomain` that is always called before? https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/cd2407f11b24a785ce6df1f67646c001b939beb3/controllers/nutanixmachine_controller.go#L484

This has the benefit of having the status be set in a single place and validation to protect against unintended changes to the Machine's spec.

yanhua121: validateMachineConfig() and validateFailureDomainRef() are only called when it needs to create the VM for the machine. While checkFailureDomainStatus() need to call for every reconcileNormal call. We can move these 4 lines of code to a separate function fetchFailureDomainObj(). But since it's just 4 lines of code, it is not much difference.

dkoshkin: @yanhua121 I think `validateFailureDomainRef()` is pretty unexpected now that it is mutating the objects, thats why I suggested to refactor to a single place.

controllers/nutanixmachine_controller.go
yanhua121: changed as suggested.

controllers/helpers.go
```diff
@@ -1172,3 +1172,29 @@ func detachVolumeGroupsFromVM(ctx context.Context, v4Client *prismclientv4.Clien
...
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+
+	return true
+}
```
thunderboltsid, line 1200: we should simplify and only check for one-way: i.e. are the subnets in failure domain present on the VM (not the other way around)

```go
// validateSubnetInclusion checks that all failure domain subnets are present in the machine's subnets.
func validateSubnetInclusion(machineSubnets, failureDomainSubnets []infrav1.NutanixResourceIdentifier, failureDomainName string) error {
	for _, fdSubnet := range failureDomainSubnets {
		found := false
		for _, machineSubnet := range machineSubnets {
			if machineSubnet.Equal(fdSubnet) {
				found = true
				break
			}
		}
		if !found {
			return fmt.Errorf("failure domain %q subnet (%s) is not present in the machine's subnets", failureDomainName, fdSubnet.DisplayString())
		}
	}
	return nil
}
```

to make message nicer, you can add a method on NutanixResourceIdentifier that takes care of adding type for you
```go
// DisplayString returns a human-readable string representation of the NutanixResourceIdentifier
// that includes both the type and value, suitable for error messages and logging.
func (nri NutanixResourceIdentifier) DisplayString() string {
	switch nri.Type {
	case NutanixIdentifierUUID:
		if nri.UUID != nil {
			return fmt.Sprintf("uuid=%q", *nri.UUID)
		}
	case NutanixIdentifierName:
		if nri.Name != nil {
			return fmt.Sprintf("name=%q", *nri.Name)
		}
	}
	return "unknown"
}
```

api/v1beta1/nutanix_types.go
```diff
@@ -92,6 +92,31 @@ func (nri NutanixResourceIdentifier) IsName() bool {
...
+	case NutanixIdentifierUUID:
+		if nri.UUID != nil && other.UUID != nil && *nri.UUID == *other.UUID {
+			return true
+		}
+		return nri.UUID == nil && other.UUID == nil
+	default:
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
yanhua121: I think it's better to use pointers for the function parameter especially for a struct object.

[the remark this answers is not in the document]
thunderboltsid: I'm ambivalent on that. Also, to make message nicer, you can add a method on NutanixResourceIdentifier that takes care of adding type for you
```go
// DisplayString returns a human-readable string representation of the NutanixResourceIdentifier
// that includes both the type and value, suitable for error messages and logging.
func (nri NutanixResourceIdentifier) DisplayString() string {
	switch nri.Type {
	case NutanixIdentifierUUID:
		if nri.UUID != nil {
			return fmt.Sprintf("uuid=%q", *nri.UUID)
		}
	case NutanixIdentifierName:
		if nri.Name != nil {
			return fmt.Sprintf("name=%q", *nri.Name)
		}
	}
	return "unknown"
}
```

controllers/nutanixmachine_controller.go
yanhua121: We can refactor the FD related code in the controller to make it more readable, in a separate PR later.

api/v1beta1/nutanix_types.go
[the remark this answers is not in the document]
yanhua121: addressed

yanhua121:
> Did you consider implementing an `EqualTo(other)` receiver (or something similar) on `NutanixResourceIdentifier`? It has a few receivers already.

@dlipovetsky Added

controllers/helpers.go
```diff
@@ -1172,3 +1172,29 @@ func detachVolumeGroupsFromVM(ctx context.Context, v4Client *prismclientv4.Clien
 
 	return nil
 }
+
+func resourceIdsEquals(s1, s2 []infrav1.NutanixResourceIdentifier) bool {
```
dkoshkin, line 1176: nit: why "s" as a var name?

dkoshkin (approved):
LGTM, just some minor comments

controllers/helpers.go
```diff
@@ -1172,3 +1172,29 @@ func detachVolumeGroupsFromVM(ctx context.Context, v4Client *prismclientv4.Clien
...
+	if s1 == nil && s2 == nil {
+		return true
+	}
+	if (s1 == nil && s2 != nil) ||
+		(s1 != nil && s2 == nil) ||
+		len(s1) != len(s2) {
```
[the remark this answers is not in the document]
yanhua121: This is to ensure neither s1 nor n2 is nil in the following lines of code to check the elements equality. Since we need to check the != nil for s1 and s2 anyway.

controllers/helpers.go
yanhua121: renamed

controllers/nutanixmachine_controller.go
```diff
@@ -514,11 +520,47 @@ func (r *NutanixMachineReconciler) reconcileNormal(rctx *nctx.MachineContext) (r
...
+	fdObj, err := r.getFailureDomainObj(rctx, *rctx.Machine.Spec.FailureDomain)
+	if err != nil {
+		return fmt.Errorf("failed to fetch the referent failure domain object %q: %w", *rctx.Machine.Spec.FailureDomain, err)
+	}
+
+	// Validate the NutanixMachine machine spec is consistent with that in the failure domain spec
+	// Note that when failure domain is used, the cluster/subnets fields of NutanixMachine spec are replaced with that in the failure domain spec,
+	// when the machine VM is created.
+	if !rctx.NutanixMachine.Spec.Cluster.EqualTo(&fdObj.Spec.PrismElementCluster) {
+		return fmt.Errorf("the nutanixmachine spec.cluster (%s) does not equal to that (%s) in the failure domain %q",
```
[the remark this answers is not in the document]
yanhua121: reformatted the error message.

controllers/nutanixmachine_controller.go
dlipovetsky: Thank you!

dlipovetsky:
I'd like the error message to report the differences between clusters _and_ subnets. Right now, if a difference between clusters is found, we don't report the differences between subnets. Here's the new output:

```
the NutanixMachine is not consistent with the referenced NutanixFailureDomain "fd-test": NutanixMachine.spec.cluster=unknown, NutanixFailureDomain.spec.prismElementCluster=name="tkk69yun0q"; NutanixMachine.spec.subnets=[], NutanixFailureDomain.spec.subnets=[tkk69yun0q]
```

I'm pushing a fixup commit.

dlipovetsky (approved):

PR #554, opened 2025-07-30
fix(api): Ensure failureDomains and controlPlaneFailureDomains are mutually exclusive

thunderboltsid:
Ensure only one of failureDomains or controlPlaneFailureDomains can be set on NutanixCluster

**How has this been tested?**
- Create a cluster with legacy failure domains
- Add controlPlaneFailureDomains on top of legacy failure domains in NutanixCluster
- Observe the error `spec: Invalid value: "object": Cannot set both 'failureDomains' and 'controlPlaneFailureDomains' fields simultaneously. Use 'controlPlaneFailureDomains' as 'failureDomains' is deprecated.`
- Remove the old legacy failure domains and see migration was successful

### DIFF
--- a/api/v1beta1/nutanixcluster_types.go
+++ b/api/v1beta1/nutanixcluster_types.go
@@ -44,6 +44,7 @@ const (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // NutanixClusterSpec defines the desired state of NutanixCluster
+// +kubebuilder:validation:XValidation:rule="!(has(self.failureDomains) && has(self.controlPlaneFailureDomains))",message="Cannot set both 'failureDomains' and 'controlPlaneFailureDomains' fields simultaneously. Use 'controlPlaneFailureDomains' as 'failureDomains' is deprecated."
 type NutanixClusterSpec struct {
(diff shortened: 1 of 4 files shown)


faiq (approved):

api/v1beta1/nutanixcluster_validate_test.go
```diff
@@ -102,3 +106,105 @@ func TestNutanixClusterOpenAPIValidationPCAddress(t *testing.T) {
 		})
 	}
 }
+
+func TestNutanixClusterFailureDomainValidation(t *testing.T) {
```
yanhua121, line 110: I think it's better to add test cases to the controller/nutanixcluster_controller_test.go, to validate setting of the FD fields (not allow to set both "failureDomains" and "controlPlaneFailureDomains" simultaneously) by creating the actual NuanixCluster object with envtest and check the expected error throw by the apiserver. See example of https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/main/controllers/nutanixfailuredomain_controller_test.go testcase "Update an NutanixFailureDomain Configuration" (L122-134).

thunderboltsid: Fair point. That is indeed a more robust test.

PR #570, opened 2025-09-18
fix: support for multiple container engine to build Makefile

abhay-nutanix:
Adding code to check for available container engine (docker/podman) and run the makefile


damdo:
@abhay-nutanix In addition to this change, could the units be made working even without a container engine? (with no docker/podman for example)?

It would be nice to be able to run `make test-unit` without the need of having container engines in CI/locally. (cc @thunderboltsid)

PR #572, opened 2025-09-18
fix: support multiple container engine

abhay-nutanix:
The changes include
- Support for multiple container engine, Docker and Podman
- Only throw error for container engine not found when the container engine is required.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ DOCKER_SOCKET = $(shell if [ "$(CONTAINER_ENGINE)" = "docker" ]; then \
                         fi)
 
 # Function to setup container engine runtime
-define setup_container_engine
+define select_container_engine
 	@if [ "$(CONTAINER_ENGINE)" = "none" ]; then \
 		echo "Error: No container engine found (docker or podman required)"; \
 		exit 1; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -10,8 +10,36 @@ LOCAL_IMAGE_REGISTRY ?= ko.local
...
+# Container engine detection (lazy evaluation - only computed when needed)
+CONTAINER_ENGINE = $(shell (command -v docker >/dev/null 2>&1 && echo docker) || \
+                           (command -v podman >/dev/null 2>&1 && echo podman) || \
+                           echo none)
+
+# Docker socket path (lazy evaluation)
+DOCKER_SOCKET = $(shell if [ "$(CONTAINER_ENGINE)" = "docker" ]; then \
+                           docker context inspect --format '{{.Endpoints.docker.Host}}'; \
+                        elif [ "$(CONTAINER_ENGINE)" = "podman" ]; then \
+                           echo "unix:///tmp/podman.sock"; \
```
thunderboltsid, line 22: we shouldn't hardcode a socket in the makefile. This value could change setup to setup. Use a runtime command to identify the correct one.
```
podman machine inspect --format '{{ .ConnectionInfo.PodmanSocket.Path }}'
```

PR #574, opened 2025-09-23
fix(e2e): added arm builds for e2e

atulv7:
Added arm capx-builds for e2e tests


dlipovetsky:
Will we backport this?

PR #586, opened 2025-10-01
refactor: removing prism-go-client/utils

abhay-nutanix:
Using "k8s.io/utils/ptr" instead of prism-go-client/utils


thunderboltsid:
> Will we backport this?

I don't see a particularly good reason to. This is being done in the main branch as we've removed the pointers utility from the prism-go-client and future release version of prism-go-client won't have it.

PR #712, opened 2026-04-13
fix: returning error if customAttributes failed to set (#703)

abhay-nutanix:
* fix:add AddVMCustomAttribute as a reconcile step

* fix: unit-tests

### DIFF
--- a/controllers/nutanixmachine_controller.go
+++ b/controllers/nutanixmachine_controller.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


thunderboltsid (approved):

atulv7 (approved):

atulv7 (approved):

adarshanand25 (approved):

yanhua121 (commented):
@abhay-nutanix 
Is this PR to fix a bug? Which jira bug ticket is for this PR? I see some issue with the fix and trying to understand the motivation of the fix.

controllers/nutanixmachine_controller.go
```diff
@@ -1055,6 +1068,34 @@ func (r *NutanixMachineReconciler) getOrCreateVM(rctx *nctx.MachineContext) (*vm
...
+func (r *NutanixMachineReconciler) addCustomAttributes(rctx *nctx.MachineContext, vm *vmmconfig.Vm) error {
+	ctx := rctx.Context
+	log := ctrl.LoggerFrom(ctx)
+	convergedClient := rctx.ConvergedClient
+
+	vmUUID := *vm.ExtId
+	vmName := *vm.Name
+	desiredAttr := vmCustomAttributePrefix4ProviderID + vmUUID
+
+	if slices.Contains(vm.CustomAttributes, desiredAttr) {
```
yanhua121, line 1082: There is an issue here in case of Metro failover when the vmUUID changed, then it will add two customAttributes with different vmUUID but with the same prefix "providerid:" to the VM. This will cause issue for the CCM to set the node's spec.providerID.

We just need to add the customAttributes with prefix "providerid:" when the VM is newly created. Once the VM already has the customAttribute with prefix "providerid:" set, don't reset it or add another one to the vm's customAttributes metadata.

yanhua121 (dismissed):

controllers/nutanixmachine_controller.go
abhay-nutanix: Fix in the PR: https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/pull/715

yanhua121: @abhay-nutanix @thunderboltsid 
I still prefer to add the "providerid:<vmUuid>" customAttribute to the VM metadata right after the VM is successfully created, as [here](https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/issue/metro-m2/controllers/nutanixmachine_controller.go#L1166-L1174). Why bother to refactor it to call in the reconcileNormal() func. This is only set once right after the VM is successfully created? It does not need to check (and set) at every reconciling cycle.

thunderboltsid: @yanhua121 this is the same reason why power on was moved out to a separate call as successive calls can fail and should be reattempted as optimistic locking isn't always guaranteed to pass

abhay-nutanix:
> @abhay-nutanix Is this PR to fix a bug? Which jira bug ticket is for this PR? I see some issue with the fix and trying to understand the motivation of the fix.

There is no ticket as such, this was part of refactoring and moving it out of createVM function. I think we have discussed the issue and the fix PR is already up. Please do let me know if there is something else you would like to discuss.

PR #715, opened 2026-04-16
fix: changing customAttributes if prefix providerID: is not present

abhay-nutanix:


controllers/nutanixmachine_controller.go
```diff
@@ -1079,7 +1079,9 @@ func (r *NutanixMachineReconciler) addCustomAttributes(rctx *nctx.MachineContext
 	vmName := *vm.Name
 	desiredAttr := vmCustomAttributePrefix4ProviderID + vmUUID
```
yanhua121: Move this line of code after the Line#1087. When the "providerid:" customAttribute is already set in the VM, we don't need to use the "desiredAttr".

yanhua121 (approved):

atulv7 (approved):

nilay-ntnx (approved):

PR #739, opened 2026-06-16
feat: add machine placement logic for metro

abhay-nutanix:
## Problem

Cluster API has no notion of Nutanix **Metro** / **MetroSite** failure domains. A Metro is a pair of
Prism Element (PE) clusters that present as a single stretched construct, and we want CAPX to:

1. Let a `Machine.spec.failureDomain` reference a metro construct (rather than a single PE).
2. Place each VM on a concrete `NutanixFailureDomain` (a single PE) such that a nodepool is spread
   **evenly across the two PEs** of the metro — for availability — while keeping the existing
   single-FD behavior untouched.
3. Tag the VM with the metadata (preferred-PE custom attribute, vHADomain category) the Nutanix
   platform needs to protect/move the VM.

Doing the even-spread part is the hard bit: it must be **deterministic and correct under concurrent
reconciles** (the machine controller runs with `concurrentReconciles > 1`) without serializing
reconciles and without a random/name-hash that can collide.

## Approach (and why)

A `Machine.spec.failureDomain` may now use a prefix:

| Prefix | Meaning |
|---|---|
| `NutanixMetro/<name>` | A `NutanixMetro` (exactly two underlying `NutanixFailureDomain`s) |
| `NutanixMetroSite/<name>` | A `NutanixMetroSite` that pins machines to a metro's preferred FD |

When resolving a machine's failure domain, the reconciler branches on these prefixes, picks a concrete
`NutanixFailureDomain`, validates the target PE is available, writes metro metadata to the VM, and (at
the cluster level) ensures a `NutanixVirtualHADomain` exists per referenced metro.

### Placement strategy

- **MetroSite — deterministic.** A machine is always placed on the site's `preferredFailureDomain`;
  the other FD is only a validation fallback. Balancing across sites is done by assigning nodepools to
  sites, not by the controller.
- **Metro — deterministic, per-group, self-healing, concurrency-safe.** `computeMetroPlacementIndex`
  performs a greedy least-count assignment:
  1. **Scoped to the machine's balancing group** (`metroPlacementGroupKey`): `MachineSet`, then
     `MachineDeployment`, then `MachinePool`, then control plane.
  2. **Already-placed siblings** (carrying the `native-failuredomain` label) seed per-FD counts, so
     placement self-heals after uneven scale-downs.
  3. **Pending siblings** (including the machine being reconciled) are sorted by name and assigned, in
     order, to the currently least-loaded FD (ties → lower index). The slot reached for *this* machine
     is its placement.

**Why it's concurrency-safe (no `concurrentReconciles=1`, no name hash):** every machine's
`NutanixMachine` exists before its reconcile runs, so all concurrent reconciles read the same sibling
set, sort it identically, and run the identical simulation — each lands in a distinct, balanced slot.
Sibling lists are read **uncached via `APIReader`** so a lagging informer cache can't make two
machines pick the same slot. Terminating machines are skipped so in-flight scale-downs free their slot.

### Balancing scope: why `MachineSet`, not `MachineDeployment`

Balancing is keyed by `MachineSet` (preferred) over `MachineDeployment`. Under the **default
surge-first rollout** (`maxSurge=1, maxUnavailable=0`), the replacement machine is created *before* the
machine it supersedes is deleted. If we balanced by `MachineDeployment`, the not-yet-deleted old
machines would keep the per-FD counts looking balanced and ties would keep resolving to the first FD,
skewing the new generation (e.g. `3-1` instead of `2-2`). Each rollout generation is its own
`MachineSet` (distinct name), so scoping to `MachineSet` lets the new generation balance independently
of the old one being torn down. In steady state a `MachineDeployment` has exactly one `MachineSet`, so
this is equivalent to `MachineDeployment` scoping.

## Known limitations / non-goals

- **Control-plane (KCP) rollouts** have no per-generation `MachineSet`, so the `MachineSet`-scoping
  that protects worker rollouts doesn't apply; a surge-first CP rollout can transiently skew
  (e.g. `2-1` on a 3-node CP). Blast radius is small and it self-heals on later placement events.
- **Scale-down** is not failure-domain-aware (CAPI delete policy isn't), so it can leave a temporarily
  uneven split; subsequent scale-ups self-heal it.
- **Only cluster-scope movement groups** are supported; nodepool-scope movement groups are a follow-up.


controllers/helpers.go
```diff
@@ -282,6 +289,31 @@ func GetPEUUID(ctx context.Context, client *v4Converged.Client, peName, peUUID *
...
+	return peCluster, nil
+}
+
+// IsPEAvailable returns whether the Prism Element cluster with the given UUID is currently available.
+func IsPEAvailable(ctx context.Context, client *v4Converged.Client, peUUID string) (bool, error) {
+	pe, err := GetPECluster(ctx, client, peUUID)
+	if err != nil {
+		return false, err
+	}
+	if pe.Config == nil || pe.Config.IsAvailable == nil {
```
thunderboltsid, line 318: Do we know of cases where Config is not present?

thunderboltsid: extrapolating, what's the infra-side contract wrt changing IsAvailable?

controllers/helpers.go
```diff
@@ -1275,3 +1307,118 @@ func getNutanixMetroObject(ctx context.Context, ctlclient client.Client, objectN
...
+		for _, movementGroup := range vhaDomain.Spec.MovementGroups {
+			for i := range movementGroup.CategoryRecoveryPlans {
+				crp := movementGroup.CategoryRecoveryPlans[i]
+				if crp.FailureDomainRef.Name != *preferredFailureDomain {
+					continue
+				}
+
+				preferredCategory := crp.Category
+				// validate the preferredCategory exists in PC
+				if _, err := getCategory(mctx.Context, mctx.ConvergedClient, preferredCategory.Key, preferredCategory.Value); err != nil {
```
thunderboltsid: @abhay-nutanix for now, keep this limited to only cluster-scope movement group. 
nodepool scope movement groups won't be supported just yet.

controllers/nutanixmachine_controller.go
```diff
@@ -723,6 +742,14 @@ func (r *NutanixMachineReconciler) validateFailureDomainSpec(rctx *nctx.MachineC
 		return err
 	}
 
+	peCluster, err := GetPECluster(rctx.Context, rctx.ConvergedClient, peUUID)
```
thunderboltsid: wouldn't we be making two calls effectively:
- ListCluster (extract UUID from matching cluster)
- GetCluster (UUID)

this seems quite wasteful

controllers/nutanixmachine_controller.go
```diff
@@ -732,6 +759,303 @@ func (r *NutanixMachineReconciler) validateFailureDomainSpec(rctx *nctx.MachineC
 	return nil
 }
 
+// getMetroFailureDomainSpec resolves a NutanixMetro failure domain to one of its two referenced
+// NutanixFailureDomains, validating PE availability and randomly selecting the preferred one.
```
thunderboltsid: note: this is no longer random with this change

controllers/nutanixmachine_controller.go
```diff
@@ -732,6 +759,303 @@ func (r *NutanixMachineReconciler) validateFailureDomainSpec(rctx *nctx.MachineC
...
+			return nil, err
+		}
+		if err = r.validateFailureDomainSpec(rctx, &remainingFd.Spec); err != nil {
+			log.Error(err, fmt.Sprintf("Both failureDomains of the NutanixMetro %s failed at validation.", metroName))
+			return nil, err
+		}
+		selectedFd = remainingFd
+	}
+
+	r.storeMetroPlacementSelection(rctx, selectedFd)
```
thunderboltsid, line 828: `storeMetroPlacementSelection` is basically setting this choice on the in-memory object.

thunderboltsid: followed the call trace, this comes from checkFailureDomainStatus right after which we call patchMachine.

controllers/nutanixcluster_controller.go
```diff
@@ -352,7 +363,147 @@ func (r *NutanixClusterReconciler) reconcileNormal(rctx *nctx.ClusterContext) (r
...
+			continue
+		}
+
+		vHADomain := &infrav1.NutanixVirtualHADomain{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: infrav1.GroupVersion.String(),
+				Kind:       infrav1.NutanixVirtualHADomainKind,
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      metroName,
```
vijayaraghavanr31: vhadomain name should include cluster name aswell. otherwise it will create conflict with  other clusters vhadomain.

controllers/nutanixmachine_controller.go
```diff
@@ -732,6 +759,303 @@ func (r *NutanixMachineReconciler) validateFailureDomainSpec(rctx *nctx.MachineC
...
+}
+
+// metroMachineGroupKeys maps each NutanixMachine name in the cluster to its nodepool key, resolved via
+// the owning CAPI Machine (which carries the nodepool labels). The NutanixMachine name equals the
+// Machine's infrastructureRef name.
+func (r *NutanixMachineReconciler) metroMachineGroupKeys(rctx *nctx.MachineContext, reader client.Reader) (map[string]string, error) {
+	machineList := &capiv1beta2.MachineList{}
+	if err := reader.List(rctx.Context, machineList,
+		client.InNamespace(rctx.NutanixCluster.Namespace),
+		client.MatchingLabels{capiv1beta2.ClusterNameLabel: rctx.Cluster.Name},
```
vijayaraghavanr31: Instead of listing all cluster Machines, should we filter only the Machines belonging to the current Machine’s nodepool? We only need the Machines within the same nodepool here.

controllers/nutanixmachine_controller.go
```diff
@@ -732,6 +759,303 @@ func (r *NutanixMachineReconciler) validateFailureDomainSpec(rctx *nctx.MachineC
...
+	if rctx.Machine != nil {
+		groupKey = metroPlacementGroupKey(rctx.Machine.Labels)
+	}
+	nmGroupKeys, err := r.metroMachineGroupKeys(rctx, reader)
+	if err != nil {
+		return 0, err
+	}
+
+	machineList := &infrav1.NutanixMachineList{}
+	if err := reader.List(rctx.Context, machineList,
```
vijayaraghavanr31, line 951: instead of listing all cluster Machines, should we filter only the Machines belonging to the current Machine’s nodepool?

controllers/helpers.go
abhay-nutanix: I don't think there is any case unless the PC is totally corrupted or missing metaData.
This is just guarding with checks even if v4 returns correct response all the time.

controllers/nutanixmachine_controller.go
abhay-nutanix: NutanixMachines lack nodepool labels, so we filter in-memory via the nodepool-scoped nmGroupKeys.

thunderboltsid:
First off: really nicely done. This is a genuinely hard problem and the design reflects that. Spreading machines evenly across two Metro failure domains without serializing reconciles is the kind of thing that's easy to get subtly wrong. The approach here is clever: seed counts from already-placed siblings, treat the rest as pending, sort by name, and run a deterministic greedy least-count simulation so every concurrent reconcile converges on the same balanced assignment assuming same dataset across reconciles. Reading directly from the APIReader to dodge a stale-cache race, scoping balancing to the nodepool via the CAPI owner labels, and skipping terminating machines so in-flight scale-downs free their slot: these are all thoughtful touches in the solution.

The way I went about reviewing this was to simulate nodepool lifecycle transitions against the algorithm and check whether the 2-FD split stayed even in the worst-case (adversarial CAPI reconcile + deletion ordering), since that's where this class of logic tends to break:
1. MD 0→4 (new nodepool): 2-2 (optimal)
2. MD 4→5: 3-2, optimal 
3. MD 5→3 (scale-down): 3-0 adversarial downscale can hit this (suboptimal but acceptable)
4. MD 3→4 (scale back up): self-heals one unit at a time (suboptimal but acceptable)
5. MD rolling upgrade (4 old → 4 new): 2-2 -> 3-1 this is the one I want to flag

Walking scenario 5 step by step: default strategy maxSurge=1, maxUnavailable=0 (surge-first), old MS balanced O1,O2@FD0 / O3,O4@FD1, adversarial deletePolicy deleting FD0 olds first:
```
start            live = O1@0 O2@0 O3@1 O4@1            counts=[2,2]
step1 surge N1   counts=[2,2] -> tie -> FD0            N1@0
   delete O1@0   live = O2@0 O3@1 O4@1 N1@0            [2,2]
step2 surge N2   counts=[2,2] -> tie -> FD0            N2@0
   delete O2@0   live = O3@1 O4@1 N1@0 N2@0            [2,2]
step3 surge N3   counts=[2,2] -> tie -> FD0            N3@0
   delete O3@1   live = O4@1 N1@0 N2@0 N3@0            [3,1]
step4 surge N4   counts=[3,1] -> argmin -> FD1         N4@1
   delete O4@1   live = N1@0 N2@0 N3@0 N4@1            [3,1]
final new MS:    N1,N2,N3 @ FD0   N4 @ FD1   ==>  3-1  (uneven)
```

This behavior is specific to the surge-first default: the same deletion order under `maxUnavailable=1, maxSurge=0` (delete-first) stays 2-2, because deleting the old before computing the replacement exposes the freed FD and self-corrects. So the algorithm is correct under one strategy (delete first, add later) and not the other (add first, delete later), which makes it a sharp edge.

thunderboltsid:
Reviewing Nit: this was hard to review for a PR of its size and complexity yet PR has an empty description and the commits are bare one-liners. I had to reverse-engineer the intended invariants, the rationale for the design (name-sorted greedy + APIReader instead of serializing reconciles), and which lifecycle transitions are guaranteed-even vs best-effort: all of which a PR description / commit message should state. Before merging, please add: a PR description (problem, approach + why, balancing scope, known limitations like the rollout case, testing done) to give reviewers as well as anyone debugging the change in future the context to match.

controllers/nutanixmachine_controller.go
```diff
@@ -1447,6 +1836,18 @@ func (r *NutanixMachineReconciler) getMachineCategoryIdentifiers(rctx *nctx.Mach
 		}
 	}
 
+	// Add the vHADomain category if the machine is configured with a NutanixMetro/NutanixMetroSite failureDomain.
+	if isNutanixMetroFailureDomain(rctx.Machine.Spec.FailureDomain) ||
+		isNutanixMetroSiteFailureDomain(rctx.Machine.Spec.FailureDomain) {
+		vhaCategory, err := getVHADomainCategory(rctx, r.Client)
+		if err != nil {
+			log.Error(err, fmt.Sprintf("Failed to get vHADomain category corresponding to the Machine's spec.failureDomain: %s", rctx.Machine.Spec.FailureDomain))
```
vijayaraghavanr31: we have return error from here , otherwise vm will be created without vha domain category

vijayaraghavanr31 (approved):
LGTM

thunderboltsid (approved):

PR #741, opened 2026-06-19
fix: remediate vulnerable transitive Go modules release 1.8

shubham2g:
This PR remediates dependency vulnerabilities identified in the BD (Black Duck) report by upgrading affected Go modules and applying targeted replace directives where transitive dependencies were pinned upstream.


.trivyignore
```diff
@@ -1,7 +1,13 @@
...
+# Fix requires v29.5.3 which uses new module paths (github.com/moby/moby/client);
 # no v28.x backport exists. Waiting on upstream cluster-api to migrate.
 # This project does not use Docker AuthZ plugins.
 CVE-2026-34040
+
+# CVE-2026-34986: go-jose v2 panic in JWE key-wrap decrypt path
+# gopkg.in/square/go-jose.v2 v2.6.0 — indirect dependency via k8s.io/apiserver / k8s.io/cluster-bootstrap.
+# v2.6.0 is the latest v2 tag; upstream fix exists only in go-jose v3.0.5/v4.1.4 lines.
+# Waiting on upstream Kubernetes/cluster-api dependency migration.
+CVE-2026-34986
```
thunderboltsid, line 13: why ignore when we can fix?

shubham2g: Hi @thunderboltsid CVE-2026-34986 is in the indirect dependency gopkg.in/square/go-jose.v2@v2.6.0, pulled through k8s.io/apiserver and k8s.io/cluster-bootstrap (v0.32.3). We verified this in CAPI 1.10.7 (used in 2.17) as well. There is no patched v2 release; fixes are only in go-jose v3/v4, which are breaking upgrades. So we are temporarily ignoring it and will remove the ignore once upstream Kubernetes/CAPI migrates.

thunderboltsid (approved):

Vedansh074 (approved):

PR #744, opened 2026-06-23
feat: controller for vhaDomain

abhay-nutanix:
Adds the `NutanixVirtualHADomain` reconciler, which manages the Prism Central
resources that back a CAPX metro / synchronous-replication HA domain. Given a
`NutanixMetro` (exactly two failure domains), the controller generates and keeps
in sync the categories, protection policy, and recovery plans needed to protect
the cluster's VMs across the two prism elements, and persists the resolved
resource UUIDs back into the `NutanixVirtualHADomain` spec/status.

## What's included

**New controller — `controllers/nutanixvirtualhadomain_controller.go`**
- Resolves the `NutanixMetro` and its two `NutanixFailureDomain`s, then for the
  cluster-scope movement group ensures, idempotently (get-or-create by name):
  - **Categories** (`k8s-vha-native-site=k8s-vha-capx-<domain>-<group>-<idx>`),
    one per failure domain.
  - **Protection policy** (`k8s-vha-capx-<domain>`) — synchronous replication
    (RPO=0) between the two PEs, tagged with both categories.
  - **Recovery plans** (`k8s-vha-capx-<domain>-<group>-<idx>`), one per failure
    domain, each with its primary-location index, network mapping, witness, and
    availability-zone configuration.
- Persists generated resource UUIDs into the spec and computes `status.Ready` /
  per-movement-group readiness.
- **Drift detection:** successfully reconciled domains are re-enqueued every 5m
  (`vhaResyncInterval`) to re-validate that the PC resources still exist, since
  PC resources cannot be watched. Surfaced via the
  `VHADomainPCResourcesValidated` condition.
- **Deletion:** finalizer-guarded. Deletion is blocked (and surfaced via
  `VHADomainSafeForDeletion=False`) until the owning `Cluster` and
  `NutanixCluster` are themselves in deletion; tolerant of owners that have
  already been garbage-collected. PC cleanup order: recovery plans → protection
  policy → categories.

**Integration changes**
- `nutanixcluster_controller.go`: on cluster deletion, deletes owned
  `NutanixVirtualHADomain`s and waits for them to be removed before dropping the
  `NutanixCluster` finalizer — guarantees PC credentials still exist for cleanup
  and prevents leaking PC resources.
- `nutanixmachine_controller.go`: VMs with a `NutanixMetro`/`NutanixMetroSite`
  failure domain are tagged at creation with the vHA category matching their
  preferred failure domain, so each VM is enrolled in the correct protection
  policy / recovery plan.
- `api/v1beta1/conditions.go`: new conditions/reasons
  (`VHADomainPCResourcesValidated`, `VHADomainSafeForDeletion`, and reasons).
- Minor: log typo fix in `pkg/client/client.go`; `go.mod`/`go.sum` updates.


vijayaraghavanr31 (approved):

controllers/nutanixvirtualhadomain_controller.go
```diff
@@ -115,10 +145,20 @@ func (r *NutanixVirtualHADomainReconciler) Reconcile(ctx context.Context, req ct
 		}
 	}()
 
-	// Add finalizer first if not set yet
+	// Handle deletion first, before resolving the owning Cluster/NutanixCluster or
+	// the Prism Central clients. Those owners may already be garbage-collected; if
+	// we required them here, the lookup would error and we would never reach the
+	// deletion handler, trapping the finalizer and blocking deletion forever.
+	if !vHADomain.DeletionTimestamp.IsZero() {
```
yanhua121: I don't think it's a right way to call reconcileDelete() before retrieving its owner Cluster/NutanixCluster, especially before adding the Finalizer. See the other reconcilers (NutanixCluster and NutanixMachine), they all calling reconcileDelete() after adding the Finalizers and getting its context including the owning objects. Please follow the same pattern as those reconcilers.

yanhua121 (changes requested):

controllers/nutanixvirtualhadomain_controller.go
```diff
@@ -174,38 +214,149 @@ func (r *NutanixVirtualHADomainReconciler) Reconcile(ctx context.Context, req ct
...
-	return ctrl.Result{}, err
+
+	// Periodically re-enqueue so the vHADomain's Prism Central resources are re-validated. PC
+	// resources cannot be watched, so this requeue is what detects out-of-band deletion/drift.
+	return ctrl.Result{RequeueAfter: vhaResyncInterval}, nil
 }
 
-func (r *NutanixVirtualHADomainReconciler) reconcileDelete(rctx *nctx.VHADomainContext) error {
-	log := ctrl.LoggerFrom(rctx.Context)
+func (r *NutanixVirtualHADomainReconciler) reconcileDelete(ctx context.Context, vHADomain *infrav1.NutanixVirtualHADomain) error {
```
yanhua121: The reconcileDelete() implementation should also be the original one.

yanhua121 (changes requested):

controllers/nutanixvirtualhadomain_controller.go
abhay-nutanix: this drops the edge-case tolerance for a genuinely orphaned vHADomain (e.g. missing cluster label or a manually-deleted CAPI Cluster while the vHADomain lingers). In that case Reconcile will error on owner lookup and requeue rather than force-cleanup — which is the same behavior as the other reconcilers.
But I will make it as other controllers.

controllers/nutanixvirtualhadomain_controller.go
```diff
@@ -219,36 +296,791 @@ func (r *NutanixVirtualHADomainReconciler) reconcileNormal(rctx *nctx.VHADomainC
...
+		if err != nil {
 			rctx.VHADomain.Status.Ready = false
 			return err
 		}
+		failureDomains = append(failureDomains, fd)
+	}
+
+	// Get or create the PC resources of the vHA domain and persist their UUIDs in the spec.
+	if err := r.ensureVHADomainPCResources(rctx, failureDomains); err != nil {
+		rctx.VHADomain.Status.Ready = false
```
yanhua121: Add this to the beginning of this reconcileNormal() func, instead of every time to set the Status.Ready to false if returning with error being not nil.

```
reconcileNormal(...) (reterr error) {
  defer func() {
    if reterr != nil {
      rctx.VHADomain.Status.Ready = false
    }
  }()
}
```

yanhua121 (commented):
Review comment

yanhua121 (approved):

PR #745, opened 2026-06-23
feat: wire PC SA API key auth method

manoj-nutanix:
- In order to use PC SA API Key with CAPX, CAPX needs to be wired to use updated prism-go-client. 

Fixes #
 [NCN-113632
](https://jira.nutanix.com/browse/NCN-113632)

- capx [infra yaml](https://github.com/nutanix-cloud-native/cluster-api-provider-nutanix/blob/645b2c556a0db5271dbdcad07e6ffdbc8e361c8b/config/manager/manager.yaml#L36-L48) - credential secret not updated to use API Key as it's legacy method which will be removed in future release. So customers relying on this method if they want to use API key - should either manually update this CAPX infra yaml or move to using `NutanixCluster.spec.credentialRef` instead.
- CCM secrets in templates/ dir not updated to use API key as current CCM release doesn't support it.

**Release note**:

```release-note

```

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -8,6 +8,7 @@
     "gnumake@latest",
     "ginkgo@latest",
     "go@latest",
+    "git@latest",
     "golangci-lint@latest",
     "gotestsum@latest",
     "kubernetes-helm@latest",
--- a/templates/topology/secret.yaml
+++ b/templates/topology/secret.yaml
@@ -7,12 +7,14 @@ stringData:
   credentials: |
     [
       {
-        "type": "basic_auth",
+        "type": "${NUTANIX_CREDENTIALS_TYPE}",
         "data": {
           "prismCentral":{
             "username": "${NUTANIX_USER}",
-            "password": "${NUTANIX_PASSWORD}"
-          }
+            "password": "${NUTANIX_PASSWORD}",
+            "apiKey": "${NUTANIX_API_KEY}"
+          },
+          "prismElements": null
         }
       }
     ]
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -101,8 +101,10 @@ jobs:
       - name: Test build
         run: devbox run -- make ${{ inputs.make-target }} LABEL_FILTERS='${{ inputs.e2e-labels }}'
         env:
+          NUTANIX_CREDENTIALS_TYPE: ${{ secrets.NUTANIX_API_KEY != '' && 'api_key' || 'basic_auth' }}
           NUTANIX_USER: ${{ secrets.NUTANIX_USER }}
           NUTANIX_PASSWORD: ${{ secrets.NUTANIX_PASSWORD }}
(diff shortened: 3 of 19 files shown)


abhay-nutanix (approved):

templates/topology/secret.yaml
```diff
@@ -7,12 +7,14 @@ stringData:
...
       {
-        "type": "basic_auth",
+        "type": "${NUTANIX_CREDENTIALS_TYPE}",
         "data": {
           "prismCentral":{
             "username": "${NUTANIX_USER}",
-            "password": "${NUTANIX_PASSWORD}"
-          }
+            "password": "${NUTANIX_PASSWORD}",
+            "apiKey": "${NUTANIX_API_KEY}"
```
thunderboltsid, line 15: username and passwords shouldn't be present when basic_auth is not used

devbox.json
```diff
@@ -8,6 +8,7 @@
     "gnumake@latest",
     "ginkgo@latest",
     "go@latest",
+    "git@latest",
```
thunderboltsid, line 11: what change requires a new dependency on git?

manoj-nutanix: few warnings pop up while starting devbox, running lint target and to perform any git push/pull/commit, we've exit devbox shell so added it -
```
✗ devbox shell                   
Info: New devbox available: 0.17.2 -> v0.17.5. Please run `devbox version update`.
Info: Ensuring packages are installed.
Warning: Devbox was unable to authenticate with the Jetify Nix cache. Some packages might be built from source.
Info: Installing the following packages to the nix store: gotestsum@latest, path:./hack/flakes#gocov-xml, clusterctl@latest, envsubst@latest, kind@latest, kubectl@latest, path:./hack/flakes#go-junit-report, path:./hack/flakes#gocov, bash@latest, fping@latest, path:./hack/flakes#go-mod-upgrade, ginkgo@latest, ko@latest, path:./hack/flakes#setup-envtest, coreutils@latest, gnumake@latest
✓ Computed the Devbox environment.
Starting a devbox shell...
warning: unhandled Platform key FamilyDisplayName
error: tool 'git' not found

(devbox) ➜ make lint
warning: unhandled Platform key FamilyDisplayName
error: tool 'git' not found
golangci-lint run -v
```

templates/topology/secret.yaml
manoj-nutanix: prism-go-client handles the cred parsing, based on  `"type": "${NUTANIX_CREDENTIALS_TYPE}",` and considers the corresponding fields only and drops others while unmarshalling so atm not a functional issue but to keep it clean, we'll need to maintain file generation flow maybe via Makefile .. will check on it

thunderboltsid (approved):

templates/topology/secret.yaml
thunderboltsid: can be handled in a separate PR.

PR #746, opened 2026-06-23
fix: NutanixVirutalHADomain crd

yanhua121:
[Metro] fix NutanixVirutalHADomain api
  - remove the unnecessary extra layer of ProtectionPolicy field in the spec
  - avoid using the map for MovementGroups, to follow the [K8s api convention](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps). 
  - add necessary CEL rules to NutanixMetro, NutanixMetroSite and NutanixVirutalHADomain spec
  - add unit tests for the Metro api


abhay-nutanix (approved):

api/v1beta1/nutanixvirtualhadomain_types.go
```diff
@@ -37,34 +37,33 @@ type NutanixVirtualHADomainSpec struct {
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=64
+	// +kubebuilder:validation:Pattern=`[a-z0-9]([-a-z0-9]*[a-z0-9])?`
+	Name string `json:"name"`
+
 	// categoryRecoveryPlans is the list of category-to-recovery-plan mappings whose
 	// member entities belong to this movement group.
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=2
```
vijayaraghavanr31, line 65: min Items should be 1, there is no need to create 2 recovery plans everytime.
cc: @abhay-nutanix

abhay-nutanix: for current implementation as it is site-wide we can accept min as 2.
We will have to change it to 1 in the future if required.

api/v1beta1/nutanixvirtualhadomain_types.go
```diff
@@ -86,13 +85,6 @@ type NutanixCategoryRecoveryPlan struct {
 	FailureDomainRef corev1.LocalObjectReference `json:"failureDomainRef"`
 }
 
-// NutanixMovementGroupStatus captures the observed state of a movement group within a virtual HA domain.
-type NutanixMovementGroupStatus struct {
-	// ready is set to true when the movement group PC resources (categories, recovery plans) are valid and ready.
-	// +kubebuilder:default=false
-	Ready bool `json:"ready"`
-}
```
vijayaraghavanr31, line 94: where are we going to capture the status of the movementgroup.

abhay-nutanix: we are not capturing it individually at this point.
The Ready status will be an overall status for a vhaDomain
Anyways it will be true only if all movementGroups are positive

abhay-nutanix (approved):

abhay-nutanix (approved):

atulv7 (approved):

api/v1beta1/nutanixvirtualhadomain_types.go
thunderboltsid: Vijay, is right, this should've stayed 1.

PR #751, opened 2026-07-09
fix: resolve NutanixMetroSite in status.failureDomain for Metro FDs

vijayaraghavanr31:
When a NutanixMachine is placed on a Metro failure domain, the status.failureDomain field previously reflected the coarse metro name (NutanixMetro/<name>). Observers had to inspect the metro.nutanix.com/native-failuredomain label to know which AHV site the node was actually running on.

This PR refines status.failureDomain to the specific NutanixMetroSite (NutanixMetroSite/<site-name>) by looking up the
NutanixMetroSite whose spec.metroRef and spec.preferredFailureDomain match the native FD label on the machine. The lookup is best-effort: if no matching site is found the metro-level name is kept unchanged.

Fixes # https://jira.nutanix.com/browse/NCN-114521

Deployed capx build with this change in nkp cluster and tested the fix:
kubectl get nutanixmachines : Failure domains pointing to NutanixMetro
beforefix

Capx with change: Deleted one of the machine/nutanixmachine: capi/capx recreated it with proper metro site reference:
afterfix

**Release note**:

```release-note

```

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -1324,6 +1324,23 @@ func getNutanixMetroSiteObject(ctx context.Context, ctlclient client.Client, obj
 	return metroSiteObj, nil
 }
 
+// findMetroSiteForNativeFD returns the name of the NutanixMetroSite in namespace whose
+// spec.metroRef.name matches metroName and spec.preferredFailureDomain.name matches nativeFDName.
+// An empty string is returned (without error) when no match is found; the caller should fall back
+// to the metro-level failure domain name.
+func findMetroSiteForNativeFD(ctx context.Context, ctlclient client.Client, namespace, metroName, nativeFDName string) (string, error) {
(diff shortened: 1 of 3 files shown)


yanhua121 (changes requested):
I don't think we need to fix this. The NutanixMachine CR's status.failureDomain need to be consistent to that of the corresponding Machine CR's spec.failureDomain. This is by design. The info of the actual native PE site that the node VM provisioned to is reflected by the label.

thunderboltsid:
> I don't think we need to fix this. The NutanixMachine CR's status.failureDomain need to be consistent to that of the corresponding Machine CR's spec.failureDomain. This is by design. The info of the actual native PE site that the node VM provisioned to is reflected by the label.

I agree, the status.failureDomain needs to be consistent to the machine's spec. If we want to express the present site granularity, consider adding another field.

PR #755, opened 2026-07-13
fix: prevent duplicate VMs for a single machine (NCN-115785)

thunderboltsid:
## Summary

CAPX can create a second VM for the same CAPI `Machine`, and can orphan a
still-running VM on delete, whenever the VM's identity cannot be resolved from
its last-known ExtId — for example after a `clusterctl move` (status not yet
repopulated on the target cluster) or after an unplanned Prism Element failover
that reassigns the VM a new ExtId. This PR closes those idempotency gaps and
tightens VM identity resolution, with unit tests throughout.

Jira: [NCN-115785](https://jira.nutanix.com/browse/NCN-115785)

## Changes

- **create-guard (`getOrCreateVM`)**: never create a VM for a `NutanixMachine`
  that already carries `spec.providerID`; requeue and retry instead.
- **identity fallback (`GetVMUUID`)**: when `status.vmUUID` is empty, fall back
  to `spec.providerID` instead of dropping to name-based lookup.
- **failover rediscovery (`FindVM`)**: when the last-known ExtId no longer
  resolves, rediscover the VM via the server-side `biosUuid` filter on PC 7.6+
  and the `providerid:<ExtId>` custom attribute via name on PC 7.5 / as a
  universal fallback.
- **disambiguation (`FindVMByName`)**: when multiple VMs share a name, prefer the
  one carrying this machine's provider-id custom attribute, otherwise adopt the
  oldest deterministically instead of dead-locking.
- **ExtId-only identity resolution (`GetVMUUID`)**: stop returning the
  guest-reported biosUUID (`Machine.Status.NodeInfo.SystemUUID`) as a lookup
  key. It shares the ExtId value only at creation; after a failover it is never
  the current ExtId, so feeding it to `VMs.Get` misses. `GetVMUUID` now consults
  only ExtId-namespace sources (`status.vmUUID`, then `spec.providerID`); the
  biosUUID keeps its legitimate role as a rediscovery filter.
- **delete-path rediscovery (`reconcileDelete`)**: before concluding a VM is
  already deleted and dropping the finalizer, attempt the same failover
  rediscovery and normalize to the VM's current ExtId — otherwise a post-failover
  delete orphans a running VM.

## Tests
Dev-PC self-hosted (clusterctl move) e2e with a TAM-provisioned PC user: SelfHostedSpec passed (1/1). A VM watcher polled PC throughout creation, pivot, and cleanup: exactly one control-plane VM and one worker VM were created (peak 2), no duplicate VM names/ExtIds were observed, the pivoted CAPX controller issued no VM creates, and teardown left no test VMs.

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -188,15 +188,16 @@ func GenerateProviderID(uuid string) string {
 //     even after the live ExtId later changes.
 //
 // It intentionally does NOT consult Machine.Status.NodeInfo.SystemUUID. That
-// field is the guest-reported (current) biosUUID, which equals the ExtId only at
-// creation time by coincidence: after an unplanned Prism Element failover the
-// ExtId changes (both versions) while the biosUUID either changes (PC 7.5) or
-// stays put (PC 7.6) - in every post-failover case it is not the current ExtId,
-// so feeding it to VMs.Get would miss. The biosUUID still matters for recovery,
-// but note that providerID is itself pinned to the original biosUUID value, so
-// that value doubles as the biosUUID filter during rediscovery (see
-// findVMByStableIdentifier) - which is why SystemUUID is never needed as a direct
-// lookup key here.
+// field is the guest-reported biosUUID, and its relationship to the live ExtId is
+// version-dependent, which makes it an unreliable VMs.Get key: after an unplanned
+// Prism Element failover PC 7.6 keeps the biosUUID pinned to the original value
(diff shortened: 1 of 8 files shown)


abhay-nutanix (commented):
Changes look good to me. Is there anything remaining for this?

thunderboltsid:
going a different way. closing this one.

PR #759, opened 2026-07-15
feat: Garbage-collect orphaned CAPX VMs

abhay-nutanix:
Adds a cluster-scoped garbage collector that deletes Nutanix VMs which CAPX created but
which are no longer tracked by any `NutanixMachine`/`Machine`. This heals VM leaks such as
the metro/vHA site-failover case in ENG-945088, where a manually powered-off worker VM is
replaced by Cluster API and its `NutanixMachine` is deleted while the peer Prism Element is
disconnected — so the delete only reaches the reachable PE, and the stale copy resurfaces
on the other PE (with no owning `NutanixMachine` left to reconcile it) once it rejoins Prism
Central.

The GC lives in the `NutanixCluster` reconciler (not the vHA controller) because the
ownership signal — the `KubernetesClusterName=<clusterName>` category CAPX stamps on every
VM — is cluster-scoped, so the same leak can occur outside metro. This makes the cleanup
generic across all clusters.

## Safety / conservatism
A VM is deleted only when **all** of the following hold:
1. it carries this cluster's ownership category (`KubernetesClusterName=<clusterName>`);
2. it carries the CAPX `providerid:` custom attribute (i.e. CAPX created it);
3. it is older than `orphanedVMGCGracePeriod` (15m) — guards against racing in-flight
   provisioning where the owning `NutanixMachine` hasn't recorded the UUID yet;
4. its UUID and name match no live `NutanixMachine`/`Machine`.
Tunables: `orphanedVMGCGracePeriod = 15m`, `orphanedVMGCInterval = 10m`.

### End-to-end on a live cluster (KIND mgmt cluster + real Prism Central)
Validated against a running workload cluster (`ns-topology/mycluster-without-topology`):
1. **Build/deploy the provider with the change**: rebuilt the manager image from the working
   tree and loaded it into the KIND management cluster (`kind load`), then restarted
   `capx-controller-manager` (avoided `make deploy`, which does `clusterctl delete
   --include-crd` and would tear down existing CRs).
2. **Create a disposable worker**: scaled the worker `MachineDeployment` to 1. Confirmed via
   the Prism v4 API that the new VM carried the ownership category
   `KubernetesClusterName=mycluster-without-topology` and the custom attribute
   `providerid:<vmUUID>`.
3. **Manufacture a real orphan** (reproduces the leak without needing an actual site-down):
   - scaled `capx-controller-manager` to 0 (so the normal delete path can't remove the VM),
   - excluded node draining on the Machine, scaled the `MachineDeployment` to 0,
   - removed the `NutanixMachine` finalizer so the `NutanixMachine` and `Machine` objects were
     deleted while the VM survived in Prism → a VM with the category + `providerid:` and no
     owning k8s objects.
4. **Run the GC**: scaled `capx-controller-manager` back to 1 and forced a `NutanixCluster`
   reconcile. (For a fast loop, `orphanedVMGCGracePeriod`/`orphanedVMGCInterval` were
   temporarily shortened, then reverted to 15m/10m.)
5. **Result**:
   - Log: `Deleting orphaned CAPX VM not tracked by any NutanixMachine`
     `vmName=mycluster-without-topology-wmd-... vmUUID=86e1be62-...`
   - Orphan VM → Prism v4 GET returns **404** (deleted).
   - Live control-plane VM → Prism v4 GET returns **200** (untouched — no false positives).
   - Cluster remained healthy; no GC errors on subsequent resyncs.

## Risk / rollout notes
- Every `NutanixCluster` now resyncs every `orphanedVMGCInterval` (10m) instead of going idle
  after `Ready`. This is intentional (needed to drive drift/orphan detection) and is a mild
  load change.
- The GC is best-effort and heavily gated (category + providerID + age + no live owner), so
  the blast radius is limited to CAPX-created, untracked VMs.

### DIFF
--- a/controllers/nutanixcluster_controller.go
+++ b/controllers/nutanixcluster_controller.go
@@ -145,7 +145,9 @@ func (r *NutanixClusterReconciler) mapNutanixFailureDomainToNutanixCluster() han
 
 // +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;update;delete
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines,verbs=get;list;watch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixclusters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixmachines,verbs=get;list;watch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixclusters/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixclusters/finalizers,verbs=update
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=nutanixfailuredomains,verbs=get;list;watch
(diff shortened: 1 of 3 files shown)


thunderboltsid:
> where a manually powered-off worker VM is replaced by Cluster API and its NutanixMachine is deleted while the peer Prism Element is disconnected 

How did NutanixMachine get deleted? We usually have a precondition of VM Delete succeeding before NutanixMachine is allowed to be deleted.

PR #760, opened 2026-07-17
feat: guard for one vha category per vm

abhay-nutanix:
CSI and NKP rely on an **implicit contract** for k8s-HA: a VM's vHADomain category uses the key
`k8s-vha-native-site`, and each VM must have **exactly one** value under that key. Because this
contract is implicit, it is easy to break unknowingly over time.

This PR hardens that contract in the CAPX machine controller and adds unit tests that fail loudly if
the contract is ever changed:

- **Reconcile-time validation** — every `NutanixMachine` reconcile of a Metro/MetroSite machine now
  verifies the VM has one and only one category with the vHADomain key `k8s-vha-native-site`.
- **Contract-guard unit tests** — new UTs assert the category is always created with the
  `k8s-vha-native-site` key, and that a node is only ever assigned a single value for that key. Any
  change to the constant or the creation/assignment logic breaks these tests so the author is alerted
  to the downstream CSI/NKP dependency.

## Motivation

The vHADomain category (key `k8s-vha-native-site`) is applied to a VM only at VM-creation time and is
never re-synced. If a VM ends up with zero or more than one such category, CSI/NKP k8s-HA behavior
breaks silently. This adds an explicit safety check on every reconcile plus regression tests around
the implicit contract.

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -1448,21 +1448,51 @@ func getVHADomainCategory(mctx *nctx.MachineContext, ctlclient client.Client) (*
 	return nil, fmt.Errorf("not found vHADomain category for NutanixMachine")
 }
 
-// countVMVHADomainCategories returns the number of categories assigned to the VM that have the
-// vHADomain key (VHADomainDefaultCategoryKey, i.e. "k8s-vha-native-site"). The VM's category
-// references only carry the category extId, so the vHADomain categories are first resolved from
-// Prism Central by key and matched against the VM's assigned category extIds.
-func countVMVHADomainCategories(ctx context.Context, client *v4Converged.Client, vm *vmmconfig.Vm) (int, error) {
-	vhaCategories, err := client.Categories.List(ctx, converged.WithFilter(fmt.Sprintf("key eq '%s'", VHADomainDefaultCategoryKey)))
+// clusterVHADomainCategoryExtIds resolves the vHADomain categories (key VHADomainDefaultCategoryKey,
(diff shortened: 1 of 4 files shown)


controllers/helpers.go
```diff
@@ -1447,3 +1447,34 @@ func getVHADomainCategory(mctx *nctx.MachineContext, ctlclient client.Client) (*
 
 	return nil, fmt.Errorf("not found vHADomain category for NutanixMachine")
 }
+
+// countVMVHADomainCategories returns the number of categories assigned to the VM that have the
+// vHADomain key (VHADomainDefaultCategoryKey, i.e. "k8s-vha-native-site"). The VM's category
+// references only carry the category extId, so the vHADomain categories are first resolved from
+// Prism Central by key and matched against the VM's assigned category extIds.
+func countVMVHADomainCategories(ctx context.Context, client *v4Converged.Client, vm *vmmconfig.Vm) (int, error) {
+	vhaCategories, err := client.Categories.List(ctx, converged.WithFilter(fmt.Sprintf("key eq '%s'", VHADomainDefaultCategoryKey)))
```
yanhua121: It's better not to retrieve all the categories from PC with the vhaDomain key "k8s-vha-native-site" for performance reason. In case the PC has hundreds of metro clusters (for the edge use cases), this is very costly. Just traverse the VM's categories (by UUID) and make sure there is only one with the vhaDomain key and also the category is included in the cluster's vhaDomain CR.

yanhua121 (changes requested):

vijayaraghavanr31 (approved):

yanhua121 (approved):

PR #761, opened 2026-07-28
feat: controller for metro scale-in

abhay-nutanix:
Adds the **Metro Scale-Down Balancer** controller (`MetroScaleDownBalancerReconciler`),
which keeps a stretched `NutanixMetro` worker `MachineSet` evenly distributed
across its two Prism Element (PE) sites during scale-down. It biases Cluster
API's (CAPI) machine-deletion order by maintaining the
`cluster.x-k8s.io/delete-machine` annotation on machines of the
over-represented site, so scale-down drains the fuller site first and the two
sites converge toward balance.

The controller is intentionally narrow: it only touches stretched-metro worker
pools, never mutates replica counts, and never deletes machines itself — it only
influences *which* machine CAPI removes when a scale-down happens.

## Motivation

A stretched-metro worker pool exposes only a single `spec.failureDomain` to
CAPI, so CAPI's `MachineSet` delete policy is **site-blind**. On scale-down it
can preferentially delete one PE site's machines, which:

- **Starves that Prism Element** of worker capacity, defeating the purpose of
  stretching the pool across two sites, and
- For **Rook-Ceph** clusters, can wedge OSD scheduling when one site loses too
  many nodes.

CAPI documents `cluster.x-k8s.io/delete-machine` as the top-priority signal on
every delete policy (annotated machines are always chosen first, regardless of
`Random`/`Newest`/`Oldest`). This controller uses that lever to steer deletions
toward the fuller site.

## Behavior

- **In scope only when:** the `MachineSet` failure domain has the
  `NutanixMetro/` prefix and the set is not being deleted. `NutanixMetroSite/`
  (single-site), non-metro, and control-plane pools are untouched — zero
  behavior change for them.
- **Victim selection:** groups live machines by site (via the
  `metro.nutanix.com/native-failuredomain` label on the owning
  `NutanixMachine`), computes `K` (pending scale-down → `liveTotal − replicas`;
  otherwise steady-state excess `= larger − smaller`), then does a greedy
  balanced pick (remove from the currently-fuller site each step). Deterministic
  tie-breaking: lexicographic site order, newest-first within a site.
- **Coexists with operators:** a private
  `metro.nutanix.com/managed-delete-machine` ownership marker means the
  controller only ever clears annotations it set itself; a manually-applied
  `delete-machine` is always respected and never cleared.
- **Never** mutates replica counts and **never** deletes machines — CAPI remains
  the sole actor that removes machines.

## RBAC

New kubebuilder markers (regenerated into `config/rbac/role.yaml`):

- `machinesets`, `machinesets/status` — `get; list; watch`
- `machines`, `machines/status` — `get; list; watch; update; patch`
- `nutanixmachines` — `get; list; watch`

Write access is needed only on `machines` (to patch annotations); everything
else is read-only.

### Manual end-to-end validation

Built the image from this branch, loaded it into a local KIND management cluster
(`capi-test`), deployed it onto `capx-controller-manager`, and exercised a live
stretched-metro worker pool (`MachineDeployment ab-ha-auto-m2-wmd2`, failure
domain `NutanixMetro/metro0`, two PE sites `fd0` / `fd1`).

After deploy, the controller registered cleanly:
`MetroScaleDownBalancer-controller` appears in the "Starting workers" list with
no RBAC errors.

Two annotations were used to verify behavior on each `Machine`:

- `cluster.x-k8s.io/delete-machine` — CAPI's delete signal.
- `metro.nutanix.com/managed-delete-machine` — the balancer's ownership marker;
  its presence proves the balancer (not CAPI's site-blind policy) chose the
  victim.

#### Test 1 — scale 6 → 2 (`K = 4`)

Start balanced `fd0=3 / fd1=3`. The balancer marked exactly 4 victims — **2 from
each site** — each with both annotations; CAPI deleted precisely those.

**Result:** `fd0=1 / fd1=1` — balanced.

#### Test 2 — scale to 2 (steady-state pre-mark consumed)

With the pool sitting at `fd0=1 / fd1=2` (one excess machine on `fd1`
pre-marked from a prior reconcile), scaling to 2 removed that pre-marked `fd1`
machine.

**Result:** `fd0=1 / fd1=1` — balanced.

#### Test 3 — full cycle scale 2 → 9 → 3 (`K = 6`)

Scale-up provisioned new VMs across both sites, landing at `fd0=5 / fd1=4` (all
9 site-labeled). Scaling to 3 marked the fuller-site machines; CAPI deleted
them.

**Result:** `fd0=1 / fd1=2` — balanced (diff 1, best possible for an odd count
of 3). The single excess machine on the fuller site retained the
`delete-machine` + `managed` markers (steady-state bias), so the next scale-down
converges to `1/1`.

| Test | Action | Expected | Observed | Pass |
|---|---|---|---|---|
| 1 | 6 → 2 (`K=4`) | 2+2 removed, remainder 1/1 | 2 from each site; `fd0=1/fd1=1` | ✅ |
| 2 | 3 → 2 | pre-marked excess removed | `fd0=1/fd1=1` | ✅ |
| 3 | 2 → 9 → 3 (`K=6`) | remainder balanced (`diff ≤ 1`) | `fd0=1/fd1=2`, excess pre-marked | ✅ |

> Note: a transient over-mark may be observed mid-flight during a large
> scale-down — the controller re-reconciles as deletions progress and converges
> to the correct set. It is not a bug.

## Non-goals / known limitation

- Does not decide *how many* machines to remove (replica counts stay owned by
  the user / autoscaler / CAPI), does not delete machines, and does not touch
  control-plane or single-site pools.
- A single scale-down *larger than the amount needed to reach balance* falls
  back to CAPI's site-blind policy for the extra victims past the balance point;
  the controller re-reconciles and re-marks after each batch, continuously
  nudging back toward balance. This is a strict improvement over the site-blind
  default and cannot re-create the one-sided collapse for normal (incremental)
  scale-downs.

### DIFF
--- a/controllers/nutanixmetro_scaledown_controller.go
+++ b/controllers/nutanixmetro_scaledown_controller.go
@@ -1,0 +1,393 @@
+/*
+Copyright 2026 Nutanix
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"sort"
+
+	infrav1 "github.com/nutanix-cloud-native/cluster-api-provider-nutanix/api/v1beta1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/ptr"
+	capiv1beta2 "sigs.k8s.io/cluster-api/api/core/v1beta2"
+	"sigs.k8s.io/cluster-api/util/patch"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// managedDeleteMachineAnnotation marks the cluster.x-k8s.io/delete-machine annotations that this
+// controller owns. It lets the balancer add/remove the CAPI delete annotation only on machines it
+// manages, so an operator's manually-set delete-machine annotation is respected (counted, never
+// cleared).
+const managedDeleteMachineAnnotation = "metro.nutanix.com/managed-delete-machine"
+
+// MetroScaleDownBalancerReconciler keeps a stretched-NutanixMetro worker MachineSet balanced across
+// its two Prism Element sites during scale-down.
+//
+// A worker pool placed on a stretched NutanixMetro failure domain exposes a single
+// spec.failureDomain to CAPI, so CAPI's MachineSet delete policy is site-blind and can delete the
+// machines of one site preferentially, starving that Prism Element (and, for Rook-Ceph clusters,
+// wedging OSD scheduling). This controller biases deletion by maintaining the
+// cluster.x-k8s.io/delete-machine annotation (documented by CAPI as top priority on every delete
+// policy) on the machines of the over-represented site, so scale-down drains the fuller site first
+// and the two sites stay balanced.
+//
+// Known limitation: marking guarantees the imbalance is corrected first. A single scale-down larger
+// than needed to reach balance will, past the balance point, fall back to CAPI's site-blind policy
+// for the extra victims; the controller re-reconciles and re-marks after each batch. This is a
+// strict improvement and cannot re-create the one-sided collapse for normal scale-downs.
+type MetroScaleDownBalancerReconciler struct {
+	client.Client
+	Scheme           *runtime.Scheme
+	controllerConfig *ControllerConfig
+}
+
+// NewMetroScaleDownBalancerReconciler creates a MetroScaleDownBalancerReconciler.
+func NewMetroScaleDownBalancerReconciler(client client.Client, scheme *runtime.Scheme, copts ...ControllerConfigOpts) (*MetroScaleDownBalancerReconciler, error) {
+	controllerConf := &ControllerConfig{}
+	for _, opt := range copts {
+		if err := opt(controllerConf); err != nil {
+			return nil, err
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


vijayaraghavanr31:
Potential race with CAPI's own scale-down for large/sudden deletions:

For a big scale-down (e.g. 19->5, 14 victims) with no pre-existing imbalance, none of the 14 victims are pre-marked by the steady-state bias. That means the balancer has to Patch all 14 machines reactively, in a sequential loop — but CAPI's MachineSet controller reacts to the same Spec.Replicas change and computes+deletes its own victim list in one pass, without waiting for anyone else:
```

deletePriorityFunc, err := getDeletePriorityFunc(ms)
	if err != nil {
		return ctrl.Result{}, err
	}
	machinesToDeleteByPriority := getMachinesToDeletePrioritized(machines, machinesToDelete, deletePriorityFunc)

	var errs []error
	machinesDeleted := []*clusterv1.Machine{}
	for i, machine := range machinesToDeleteByPriority {
		log := log.WithValues("Machine", klog.KObj(machine))
		if machine.GetDeletionTimestamp().IsZero() {
			if err := r.Client.Delete(ctx, machine); err != nil {

```

machines here is a snapshot fetched at the start of that reconcile — CAPI decides the full victim list once and deletes them all in that same pass, using whatever annotation state exists at that instant. It doesn't retry or wait for late-arriving annotations.

So if CAPI's reconcile fetches/decides before our balancer finishes patching all 14, CAPI falls back to its default (Random) policy for the unannotated remainder — deleting from the wrong site, with no way to undo it after Delete is called.

This is different from the "overshoot past balance point" limitation already called out — that one self-corrects over subsequent reconciles; this is a one-shot race for reaching balance in the first place. Worth addressing (e.g. mark victims synchronously/atomically before the replica change is allowed to propagate) or explicitly documenting as a known limitation for large single-step scale-downs.

@thunderboltsid wdyt?

controllers/nutanixmetro_scaledown_controller.go
```diff
@@ -0,0 +1,393 @@
...
+// policy) on the machines of the over-represented site, so scale-down drains the fuller site first
+// and the two sites stay balanced.
+//
+// Known limitation: marking guarantees the imbalance is corrected first. A single scale-down larger
+// than needed to reach balance will, past the balance point, fall back to CAPI's site-blind policy
+// for the extra victims; the controller re-reconciles and re-marks after each batch. This is a
+// strict improvement and cannot re-create the one-sided collapse for normal scale-downs.
+type MetroScaleDownBalancerReconciler struct {
+	client.Client
+	Scheme           *runtime.Scheme
```
vijayaraghavanr31, line 60: scheme - are we keeping this for consistency with siblings?

vijayaraghavanr31 (approved):

Vedansh074 (approved):
LGTM

PR #764, opened 2026-08-07
ci(NCN-116738): re-add CSI 3.0 e2e tests and bump driver to 3.7.0

thunderboltsid:
## Summary
- #720 intended to remove only CSI 2.0 e2e coverage, but deleted `test/e2e/csi_test.go` in its entirety, which also removed the CSI 3.0 specs and silently zeroed out the `nutanix-storage-test` e2e job.
- Restore the CSI 3.0 `Describe`/`It` blocks in `test/e2e/csi_test.go` (dropping the CSI 2.0-only spec and its now-unused Prism Element secret fields, since CSI 3.0 provisions its secret via the CRS template using PC credentials, not in Go code).
- Bump the `nutanix-csi-storage` chart used to render `templates/csi3` from 3.3.4 to 3.7.0 (`scripts/csi3_nutanix_update.sh`), and regenerate `templates/csi3/nutanix-csi-storage.yaml` and `templates/cluster-template-csi3.yaml` accordingly.

## Test plan
- [x] `go build -tags e2e ./test/e2e/...`
- [x] `go vet -tags e2e ./test/e2e/...`
- [x] `golangci-lint run --build-tags e2e ./test/e2e/...`
- [x] CI: `nutanix-storage-test` e2e job runs the two CSI 3.0 specs against CSI driver 3.7.0

### DIFF
--- a/scripts/csi3_nutanix_update.sh
+++ b/scripts/csi3_nutanix_update.sh
@@ -42,3 +42,14 @@ ex templates/csi3/nutanix-csi-storage.yaml <<EOF
 %s/--http-endpoint=:9812/--metrics-address=:9812/
 xit
 EOF
+
+# Fix 3: Raise nutanix-storage-precheck-job's backoffLimit from 0 to 10
+# This Job is a Helm pre-install hook, but we apply the chart output directly via
+# kustomize/CRS instead of `helm install`, so Kubernetes has no hook ordering and the
+# Job races the CSI controller against a freshly-provisioned node. A backoffLimit of 0
+# means a single transient failure (e.g. CoreDNS not ready yet) permanently starves the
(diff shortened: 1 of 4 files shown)


Vedansh074 (approved):

vijayaraghavanr31 (approved):

thunderboltsid:
note: projects scope tests are expected to fail as the upstream doesn't have categories that are project scope. Once public v4 sdk is updated, we'll do a full backsync from internal capx. 

```
Failed to perform the operation 'Create VM' on the VM with UUID 'EMPTY' and the project with UUID
'16a9a086-3967-4465-ab40-ba6a6608daac' because the project association check failed. The VM cannot
be associated with entity 'Category' and with UUID list '073e2b2f-dbf1-435b-4e21-d4a81cb75955'.
```

PR #765, opened 2026-08-07
ci(e2e): reserve control plane endpoint IPs via Prism Central IPAM

thunderboltsid:
## Summary

Replaces the current e2e control-plane-endpoint-IP allocation — an `fping`/`shuf` ping-sweep over a static range configured via GitHub Actions variables — with dynamic reservation against the target subnet's own IPAM pool, using Prism Central's Networking v4 API (`Subnets.ReserveIpsBySubnetId`/`UnreserveIpsBySubnetId`). This mirrors how `cluster-api-runtime-extensions-nutanix` handles the same problem in its e2e suite.

The old mechanism was racy: it was a snapshot-in-time ICMP check with no actual reservation, run independently by every concurrent CI job (the ~8-way `build-dev.yaml` matrix plus periodic conformance workflows) against the same shared range, with no coordination between them. Prism Central owns the reservation atomically now, so concurrent jobs can never be handed the same address.

* `test/e2e/subnet_ip_reservation.go` (new) — `reserveSubnetIP`/`unreserveSubnetIP`/`waitForNetworkingTask`, wrapping the PC v4 Subnets/Tasks services CAPX already depends on.
* `test/e2e/test_helpers.go` — `testHelper.reserveControlPlaneEndpointIP`, resolving PE/subnet UUIDs via the existing `controllers.GetPEUUID`/`GetSubnetUUID` helpers.
* `test/e2e/e2e_suite_test.go` — reserves `CONTROL_PLANE_ENDPOINT_IP` and `CONTROL_PLANE_ENDPOINT_IP_WORKLOAD_CLUSTER` once per Ginkgo ParallelNode in `SynchronizedBeforeSuite`, releases them in `SynchronizedAfterSuite` (runs even on spec failure).
* `test/e2e/categories_test.go` — simplified now that the workload-cluster IP is always pre-reserved.
* `.github/workflows/e2e.yaml`, `Makefile` — removed the `nutanix-cp-endpoint-ip` target and the range pre-computation steps/env wiring; no longer needed.
* `test/e2e/config/nutanix.yaml`, `docs/developer_workflow.md` — documentation updates.

## Test plan

- [x] `go build ./...` and `go build -tags e2e ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run --build-tags e2e ./test/e2e/...` (0 issues)
- [x] Compiled the e2e Ginkgo test binary (`go test -tags e2e -c ./test/e2e/`)
- [x] Live-verified the reserve/unreserve round trip against a real Prism Central (`pc.dev.nkp.sh`, subnet `vlan173`): resolved PE/subnet UUIDs, reserved a fresh IP via `Subnets.ReserveIpsBySubnetId`, then released it via `Subnets.UnreserveIpsBySubnetId` — both completed successfully end to end, including the async task-polling and completion-details parsing.
- [x] Confirmed the target subnet has a real managed IPAM pool PC can reserve from (502 addresses total). Note: the pool was initially full of reservations left over from the old never-releasing approach, which had to be cleared by hand before a fresh reservation would succeed — a good illustration of the exact problem this PR fixes going forward.

### DIFF
--- a/test/e2e/subnet_ip_reservation.go
+++ b/test/e2e/subnet_ip_reservation.go
@@ -1,0 +1,169 @@
+//go:build e2e
+
+/*
+Copyright 2026 Nutanix
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"time"
+
+	v4Converged "github.com/nutanix-cloud-native/prism-go-client/converged/v4"
+	networkingcommonapi "github.com/nutanix/ntnx-api-golang-clients/networking-go-client/v4/models/common/v1/config"
+	networkingapi "github.com/nutanix/ntnx-api-golang-clients/networking-go-client/v4/models/networking/v4/config"
+	prismcommonapi "github.com/nutanix/ntnx-api-golang-clients/prism-go-client/v4/models/common/v1/config"
+	prismapi "github.com/nutanix/ntnx-api-golang-clients/prism-go-client/v4/models/prism/v4/config"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/utils/ptr"
+)
+
+const (
+	ipReservationTaskPollInterval = 2 * time.Second
+	ipReservationTaskTimeout      = 2 * time.Minute
+)
+
+// reservedIPsCompletionDetails mirrors the "reserved_ips" completion detail Prism Central
+// attaches to a successful ReserveIpsBySubnetId task.
+type reservedIPsCompletionDetails struct {
+	ReservedIPs []string `json:"reserved_ips"`
+}
+
+// reserveSubnetIP asks Prism Central to atomically reserve a single free IP address from the
+// given subnet's own IPAM pool, and returns a function that releases it again. Because Prism
+// Central owns the reservation, concurrent callers (e.g. separate e2e CI jobs targeting the same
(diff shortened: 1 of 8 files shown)


abhay-nutanix (approved):

test/e2e/subnet_ip_reservation.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to wait for IP reservation task: %w", err)
+	}
+	if len(completionDetails) == 0 {
+		return "", nil, fmt.Errorf("reserve-IP task for subnet %s completed with no details", subnetUUID)
+	}
+
+	// The reserved address comes back as a JSON-encoded string in the first completion detail
+	// value, e.g. `"{\"reserved_ips\":[\"10.0.0.5\"]}"`.
+	marshaledValue, err := json.Marshal(completionDetails[0].Value)
```
adarshanand25, line 82: Can we avoid doing json.Marshal + strconv.Unquote round trip here? 
Since Value is already coming as JSON-encoded string, we should be able to unmarshal it directly into reservedIPsCompletionDetails.

thunderboltsid: Good catch, done in 536d3c6. Value.GetValue() already returns the raw JSON string for the String-typed KVPair, so I dropped the marshal/unquote round trip and unmarshal it directly now. Verified live against Prism Central that reserve/unreserve still round-trip correctly with the simplified parsing.

test/e2e/subnet_ip_reservation.go
```diff
@@ -0,0 +1,170 @@
...
+	var reserved reservedIPsCompletionDetails
+	if err := json.Unmarshal([]byte(rawValue), &reserved); err != nil {
+		return "", nil, fmt.Errorf("failed to unmarshal reserve-IP completion details %s: %w", rawValue, err)
+	}
+	if len(reserved.ReservedIPs) == 0 {
+		return "", nil, fmt.Errorf("reserve-IP task for subnet %s reserved no addresses", subnetUUID)
+	}
+
+	reservedIP := reserved.ReservedIPs[0]
+	return reservedIP, func() error {
```
adarshanand25, line 101: You reserve the IP using ctx, then return a cleanup function[unreserveSubnetIP] that also depends on that same ctx; if ctx expires before cleanup, cleanup may fail. just want to understand this scenario

adarshanand25 (approved):
LGTM

PR #778, opened 2026-08-27
fix: delete recovered VM after Metro UPFO instead of orphaning it

abhay-nutanix:

After Metro unplanned failover (UPFO), the Machine still records the **original** VM UUID (`systemUUID` / `status.vmUUID`). The recovered AHV VM on the surviving site keeps the **same name** and gets a **new UUID**. The original leftover is DR-decoupled (`role=kDecoupled`).

Older `reconcileDelete` only looked up that recorded UUID:

1. While the failed site is down, delete of the decoupled UUID fails → `NutanixMachine` finalizer sticks → KCP will not create a replacement control-plane Machine.
2. When the site returns, DR deletes the decoupled leftover. CAPX sees the recorded UUID as gone, drops the finalizer, and **never deletes the recovered VM** → orphan.

This change:

- Never deletes a VM whose Prism Central DR config `role` is `kDecoupled` (DR owns it).
- Finds the live recovered VM **by Machine name**, skipping decoupled UUIDs, and deletes **that** UUID.
- If the recorded UUID is already gone (DR cleaned the leftover), still looks up by name so the migrated VM is not left behind.
- Persists `capx.nutanix.com/recovered-vm-uuid` so wait-vs-done can be distinguished while the original UUID is still decoupled.
- Same path for control-plane and worker `NutanixMachine` objects.

### Unit tests

```bash
go test ./controllers -count=1 -timeout 120s \
  -run 'TestIsVMDecoupled|TestFindNonDecoupledVMByName|TestNutanixMachineReconciler_ReconcileDelete'
```

New coverage:

| Test | What it asserts |
|---|---|
| `TestIsVMDecoupled` | `role=kDecoupled` is decoupled; other roles are not; groups API errors surface |
| `TestFindNonDecoupledVMByName` | skips decoupled UUID, returns the live recovered VM; nil when only decoupled VMs exist |
| ReconcileDelete: wait when recovered is missing | finalizer kept, `skipped-decoupled-vm-uuid` set |
| ReconcileDelete: delete recovered instead of original | `DeleteAsync` on recovered UUID, not the decoupled original |
| ReconcileDelete: finalizers after recovered is gone | finalizers dropped |
| ReconcileDelete: DR already removed the original | name lookup finds migrated VM and deletes **that** UUID |

Existing ReconcileDelete cases (no v3 client) still take the UUID-only path.

### Live cluster (Metro UPFO + MHC remediation)

Built from `5420d945` and rolled out via `CAPIStack/capi-stack` on management cluster `abhay-mgmt`:

`harbor.eng.nutanix.com/abhay-dev/cluster-api-provider-nutanix:e2e-5420d945a22c95606ed72bc424eb9ceacd54e64a@sha256:3ef0161cca7cf118acaf26133ea9e5471ccc0c81e58c2e4e13779d9896ac76a8`

Remediation of control-plane Machine `kommander/abhay-mgmt-7p5fr-p26n6`. CAPX logs (`capx-system`, `14:53:08Z`–`14:53:15Z`):

1. Recorded UUID `ef396be5-5ac3-4cfa-7827-61c6558bdcb9` was **gone** → name lookup (the site-back / DR-cleaned-original path).
2. Skipped decoupled leftover `c3b3c8c1-fe37-42ad-5a3c-f6b75ba4c54f` (not deleted).
3. Deleted live recovered VM `358e14da-94a6-4908-7ab0-0fe841d37784`.
4. After that UUID was gone, removed finalizers. `NutanixMachine` and `Machine` were both gone by `14:53:55Z`.

Without this change, CAPX would have dropped the finalizer when `ef396be5` was missing and left `358e14da` running unmanaged.

A second CP Machine on the same cluster (`abhay-mgmt-7p5fr-pkt89`) took the **normal** delete path: deleted the recorded UUID, name lookup found no recovered leftover, finalizers removed. Confirms non-UPFO delete is unchanged.

- Decoupled is `role == kDecoupled` only (groups `entity_dr_config`). Stretch state is not used.
- `capx.nutanix.com/recovered-vm-uuid` is delete-only bookkeeping (same pattern as `metro.nutanix.com/active-placement-pe`). A boolean is not enough for a second failover.
- Wait-for-recovered returns `RequeueAfter` with **no error** so the patch helper persists annotations.
- If `NutanixClient` / v3 is missing, delete still uses the recorded UUID only (legacy).
- Design: `docs/architecture/metro-upfo-vm-delete.md`. KB: `docs/kb/metro-upfo-control-plane-remediation-deadlock.md`.

**Release note**:

```release-note
After Metro UPFO, NutanixMachine delete skips DR-decoupled leftover VMs and deletes the recovered VM (same name, new UUID) instead of leaving it orphaned.
```

### DIFF
--- a/controllers/helpers.go
+++ b/controllers/helpers.go
@@ -85,6 +85,17 @@ const (
 
 	vmCustomAttributePrefix4MetroPreferredPE        = "metro-preferred-pe:"
 	vmCustomAttributePrefix4MetroNodeGroupNameLabel = "metro-node-group-name:"
+
+	drConfigEntityType    = "entity_dr_config"
+	drRoleDecoupled       = "kDecoupled"
+	drConfigRoleAttribute = "role"
+
+	// skippedDecoupledVMUUIDAnnotation records the Machine's original VM UUID when
+	// delete skipped it because DR marked the VM decoupled (Metro UPFO).
+	skippedDecoupledVMUUIDAnnotation = "capx.nutanix.com/skipped-decoupled-vm-uuid"
+	// recoveredVMUUIDAnnotation is the non-decoupled recovered VM CAPX will delete
+	// instead of the decoupled original.
+	recoveredVMUUIDAnnotation = "capx.nutanix.com/recovered-vm-uuid"
 )
 
 type StorageContainerIntentResponse struct {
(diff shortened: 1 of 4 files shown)


controllers/nutanixmachine_controller.go
```diff
@@ -442,10 +453,100 @@ func (r *NutanixMachineReconciler) reconcileDelete(rctx *nctx.MachineContext) (r
...
+func (r *NutanixMachineReconciler) vmToDelete(rctx *nctx.MachineContext, recordedUUID, vmName string) (*vmmconfig.Vm, bool, error) {
+	ctx := rctx.Context
+	log := ctrl.LoggerFrom(ctx)
+	v3Client := nutanixV3Service(rctx.NutanixClient)
+	if v3Client == nil {
+		vm, err := FindVMByUUID(ctx, rctx.ConvergedClient, recordedUUID)
+		return vm, false, err
+	}
+
+	decoupled, err := isVMDecoupled(ctx, v3Client, recordedUUID)
```
vijayaraghavanr31, line 484: nit: @abhay-nutanix  `isVMDecoupled` (v3 Groups `entity_dr_config`) runs on every `reconcileDelete` whenever a v3 client exists, including non-metro clusters.

On the success path behaviour is the same (`kDecoupled` is false -> UUID delete as before). 
The failure path is not: a groups timeout/RBAC error now blocks delete of a normal Machine and holds the finalizer. `kDecoupled` is a Metro DR signal and is not applicable to other cluster types.

Please gate this on metro/UPFO — e.g. metro failure domain or only after we have already observed `kDecoupled` / `skipped-decoupled-vm-uuid`. 
Non-metro delete should stay UUID-only, including when Groups is down.

 we can fix it in the followup pr

vijayaraghavanr31 (approved):
LGTM

controllers/helpers.go
```diff
@@ -254,6 +265,148 @@ func FindVMByName(ctx context.Context, client *v4Converged.Client, vmName string
 	return FindVMByUUID(ctx, client, *vms[0].ExtId)
 }
 
+func nutanixV3Service(client *prismclientv3.Client) prismclientv3.Service {
+	if client == nil {
+		return nil
+	}
+	return client.V3
+}
```
thunderboltsid: this seems superfluous.

controllers/helpers.go
```diff
@@ -254,6 +265,148 @@ func FindVMByName(ctx context.Context, client *v4Converged.Client, vmName string
...
+			continue
+		}
+		for _, entity := range group.EntityResults {
+			if groupsFieldValue(entity, drConfigRoleAttribute) == drRoleDecoupled {
+				return true, nil
+			}
+		}
+	}
+	return false, nil
+}
```
thunderboltsid, line 345: why are we adding a v3 groups call?

thunderboltsid: There is no, and as per policy will not be in future either, a replacement for v3 groups APIs.

abhay-nutanix: Because we don't have a v4 equivalent. CSI is also using the same

controllers/helpers.go
abhay-nutanix: ok I will move it to generic V3 calls.

controllers/helpers.go
thunderboltsid: v4 groups API will never be there as per policy.

thunderboltsid: what's the v4 migration plan for getting rid of this particular call here?

controllers/nutanixmachine_controller.go
```diff
@@ -346,7 +346,7 @@ func (r *NutanixMachineReconciler) reconcileDelete(rctx *nctx.MachineContext) (r
 		return reconcile.Result{}, nil
 	}
 
-	vm, err := FindVMByUUID(ctx, convergedClient, vmUUID)
+	vm, waitForRecovered, err := r.vmToDelete(rctx, vmUUID, vmName)
```
thunderboltsid, line 349: If this logic relies on a v3 groups API call, CAPX cannot have it on the delete hot path for any project-scope cluster.  The v3 Groups APIs were not modified to support Projects 2.0 boundaries. Because they do not recognize project-related scopes, I'd expect IAM denies the requests.

thunderboltsid: Coming to think of it, this would actually apply to any least privilege user (projects or otherwise), right? cc @manoj-nutanix

abhay-nutanix: Gated the call only for metro and metroSite based machines.

controllers/helpers.go
abhay-nutanix: Exploring the timelines with the team.

thunderboltsid: Thanks, please keep on top of this.

controllers/nutanixmachine_controller.go
thunderboltsid: Thanks, that helps derisk rest of cluster lifecycle.

thunderboltsid:
Thanks for the changes. Please work out a plan to offboard these v3 calls.

thunderboltsid (approved):

PR #782, opened 2026-09-10
fix: fail fast if category value exceeds 64

abhay-nutanix:
Name generation now fails immediately if the result is longer than 64 characters. Nothing is hashed or truncated.

That applies to category values. 

Short names are unchanged. Shorten the cluster or metro name so k8s-vha-capx-{vhadomain}-{group}-{idx} stays within 64.

### DIFF
--- a/controllers/nutanixvirtualhadomain_controller.go
+++ b/controllers/nutanixvirtualhadomain_controller.go
@@ -56,6 +56,9 @@ const (
 	// vhaDefaultMovementGroup is the name of the single movement group generated for a vHA domain.
 	vhaDefaultMovementGroup = "default"
 
+	// prismNameMaxLen is Prism Central's maximum length for category values.
+	prismNameMaxLen = 64
+
 	// vhaResyncInterval is how often a successfully-reconciled vHADomain is re-enqueued so that its
 	// Prism Central resources (categories, protection policy, recovery plans) are re-validated.
 	// Those resources live in Prism Central and cannot be watched, so periodic requeue is used to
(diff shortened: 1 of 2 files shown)


controllers/nutanixvirtualhadomain_controller.go
```diff
@@ -616,8 +619,12 @@ func (r *NutanixVirtualHADomainReconciler) validateRecoveryPlanExists(
 // vhaCategoryValue returns the category value generated for a vHA domain's
 // movement group at the given prism-element index. The category key is shared
 // across all clusters (VHADomainDefaultCategoryKey).
-func vhaCategoryValue(vHADomainName, group string, idx int) string {
-	return fmt.Sprintf("k8s-vha-capx-%s-%s-%d", vHADomainName, group, idx)
+func vhaCategoryValue(vHADomainName, group string, idx int) (string, error) {
+	value := fmt.Sprintf("k8s-vha-capx-%s-%s-%d", vHADomainName, group, idx)
+	if len(value) > prismNameMaxLen {
```
vijayaraghavanr31, line 624: Returning this error will keep the vHA domain reconciler retrying until the name is shortened, since the spec does not change.

That's expected here as a last-line CAPX guard. Offline with @abhay-nutanix  CAREN will add a preflight for this 64-char category-value limit (`k8s-vha-capx-{vhadomain}-{group}-{idx}`) so invalid names fail admission and never hit this loop.

vijayaraghavanr31 (approved):

Vedansh074 (approved):

PR #783, opened 2026-09-16
fix: add more logging for task errors

abhay-nutanix:
## Summary
When a Prism parent task fails, CAPX often only surfaces a generic parent-task error. This change waits for converged operations and, on failure, collects error messages from failed child subtasks so machine (and related) status includes the actionable infrastructure reason (for example DHCP pool exhaustion).

- Enrich wait errors with failed subtask messages (list failed children; Get parent/subtasks only if that list call errors)
- Deduplicate `ErrorMessages` and `LegacyErrorMessage` when Prism Central fills both with the same string
- Log parent and subtask failure details for operators


controllers/task_errors.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	visited[parentTaskUUID] = struct{}{}
+
+	log := ctrl.LoggerFrom(ctx)
+	children, err := listFailedChildTasks(ctx, client, parentTaskUUID)
+	if err != nil {
+		log.Error(err, "failed to list Prism subtasks while collecting failure details", "parentTaskUUID", parentTaskUUID)
+	}
+	if len(children) == 0 {
+		children = failedSubtasksFromParentGet(ctx, client, parentTaskUUID)
```
vijayaraghavanr31, line 83: only use the Get fallback when listFailedChildTasks errors, not when it returns []

controllers/task_errors.go
```diff
@@ -0,0 +1,173 @@
...
+	op := taskOperation(task)
+
+	var parts []string
+	if task != nil {
+		for _, msg := range task.ErrorMessages {
+			if msg.Message != nil && *msg.Message != "" {
+				parts = append(parts, *msg.Message)
+			}
+		}
+		if legacy := ptr.Deref(task.LegacyErrorMessage, ""); legacy != "" {
```
vijayaraghavanr31, line 163: if pc fills both ErrorMessages and LegacyErrorMessage with the same string, both are joined??

vijayaraghavanr31:
@abhay-nutanix pls add description

vijayaraghavanr31 (changes requested):

controllers/task_errors.go
abhay-nutanix: Fixed: the parent-task Get fallback now runs only when listFailedChildTasks returns an error, not when it returns an empty list.

controllers/task_errors.go
abhay-nutanix: Fixed: ErrorMessages and LegacyErrorMessage are now appended uniquely, so the same string from PC is not joined twice.

vijayaraghavanr31 (approved):

controllers/nutanixmachine_controller.go
```diff
@@ -1862,13 +1852,43 @@ func (r *NutanixMachineReconciler) getOrCreateVM(rctx *nctx.MachineContext) (*vm
...
+	convergedClient := rctx.ConvergedClient
+	createOp, err := convergedClient.VMs.CreateAsync(ctx, vm)
+	if err != nil {
+		return nil, vmCreateFailure(rctx, vmName, err)
+	}
+	createdVMs, err := waitForConvergedOperation(ctx, convergedClient, createOp)
+	if err != nil {
+		return nil, vmCreateFailure(rctx, vmName, err)
+	}
+	if len(createdVMs) == 0 || createdVMs[0] == nil {
```
thunderboltsid: in which case would we expect this to be not exactly 1?

abhay-nutanix: We don’t expect a successful VM create to return anything other than a single VM. Operation.Wait is generic and returns []*T, so this is a defensive nil/empty check so we don’t panic on createdVMs[0] if the task completes with err == nil but no entity in the result (empty list or a nil slot).

I haven’t seen Prism return multiple VMs from a single Create, so treating “not exactly 1” as failure is reasonable. I can tighten this to:
`if len(createdVMs) != 1 || createdVMs[0] == nil {`
and include the actual count in the error.

thunderboltsid: Yes, please tighten this.

thunderboltsid (approved):

PR #784, opened 2026-09-16
fix: compare metro subnets by VLAN profile, not name

vijayaraghavanr31:
## Summary
- During Metro recovery placement, `checkFailureDomainStatus` compared `NutanixMachine.spec.subnets` to the native failure domain by subnet name/UUID.
- That fails when each PE has its own subnet object (`Vlan-041-site-01` vs `Vlan-041-site-02`) even though they are the same stretched L2.
- For `NutanixMetro` / `NutanixMetroSite` machines, resolve both subnet lists from Prism and compare **profile** (layer, VLAN ID/VNI, CIDR). Non-metro FDs still use identifier equality.

## Test plan
- [x] Unit tests for matching VLAN/CIDR across differently named subnets
- [x] Unit tests for mismatch when VLAN IDs differ
- [x] `make unit-test`
- [x] Metro failover: scale the failed-over site node pool by 1 and confirm the new machine reconciles
- [ ] Confirm a wrong VLAN on the paired site still fails the consistency check

testing logs: 
[capx_validation_logs.logs.txt](https://github.com/user-attachments/files/32333214/capx_validation_logs.logs.txt)

PR #785, opened 2026-09-17
fix: find VM flow after UPFO for metro

abhay-nutanix:
Summary

- After Metro/MetroSite UPFO, getOrCreateVM can record SystemUUID/providerID while Prism’s live ExtId has changed. FindVM did a v4 GET on that UUID, 404’d, and failed with expected to be present without listing by name — so reconcile never adopted the migrated VM.
- For Metro and MetroSite only (useMetroDRDeletePath), a UUID 404 now lists by Machine name (then NutanixMachine name). A hit is treated as the existing VM; an empty list still returns expected to be present so we do not create a duplicate.
- Non-metro create stays UUID-only on 404.


vijayaraghavanr31 (approved):
As discussed, please raise a follow-up PR for the BIOS UUID changes. This should take care of updating the ExtID reference after adoption.

Please ensure that both the Machine and NutanixMachine UUID references are synced with the live VM UUID/ExtID after adoption, so that the stale UUID is not retained.

thunderboltsid (approved):

thunderboltsid:
Folks, we really need to prioritize the sync-back PR once #783 #784 and #785 merge. Here's the plan:
* merge these three
* create a sync PR
* update the syncback PR and get that merged

